PR #15, opened 2023-07-18
Wordpress example project

MurruAlessio:
materials for wordpress installation


peppelinux (changes requested):
as mentioned during the meeting, we cannot take a commit of 2200 files

we need a detailed documentation of how to setup and configure the example wordpress project

and a docker compose that builds the wordpress dockerfile, with configuration, html/php and SQL, to be setup in the first init of the compose

peppelinux:
Closed, since we cannot take more than 2k of php files in this project

PR #35, opened 2023-07-26
test: dpop

salvatorelaiso:
While testing the property `is_valid` of `DPoPVerifier`, I found out it can fail on some invalid inputs.
Specifically, you can find some commented out test cases which causes the aforementioned property to raise different errors on not-valid inputs. Do you prefer to address those cases, for example, by returning `False` if any of the underlying checks raise an error, or we keep it as it is?


peppelinux:
is_valid MUST return a boolean

please specify it if not already defined in the property/function definition

in the test we have to assert true or false, depending by the scope of the test

peppelinux:
we must tests situation where the dpop is invalid or with a wrong schema, the tests we need wants to cover also the security aspects and not just the assertion to wellknown parameter, that we have used in the tests

pyeudiw/tests/oauth2/test_dpop.py
```diff
@@ -65,18 +65,40 @@ def wia_jws(jwshelper):
 
 def test_create_validate_dpop_http_headers(wia_jws, private_jwk=PRIVATE_JWK):
     # create
-    unpad_jwt_header(wia_jws)
+    header = unpad_jwt_header(wia_jws)
+    assert header
+    assert header["trust_chain"] == []
+    assert header["x5c"] == []
```
peppelinux, line 71:
```suggestion
    assert isinstance(header["trust_chain"], list)   
    assert isinstance(header["x5c"], list)
```

pyeudiw/tests/oauth2/test_dpop.py
```diff
@@ -65,18 +65,40 @@ def wia_jws(jwshelper):
...
+    assert header["alg"]
+    assert header["kid"]
+    
     payload = unpad_jwt_payload(wia_jws)
```
peppelinux, line 75: this is a test for the Wallet Instance Attestation while in this unit test we aims to test the DPoP proof

since you have did this test for the WIA, please move it to another test, and please in this unit test just test the DPoP proof, public key and ath binding

pyeudiw/tests/oauth2/test_dpop.py
```diff
@@ -65,18 +65,40 @@ def wia_jws(jwshelper):
...
+   
     new_dpop = DPoPIssuer(
         htu='https://example.org/redirect',
         token=wia_jws,
         private_jwk=private_jwk
     )
     proof = new_dpop.proof
-
-    # TODO assertions
+    assert proof
```
peppelinux, line 101: assert proof alone is not enough

please do tests of the ath hash over the WIA

peppelinux (changes requested):

pyeudiw/oauth2/dpop.py
```diff
@@ -81,12 +80,24 @@ def __init__(
...
+            raise ValueError("DPoP proof is not a valid JWT")
         self.proof = http_header_dpop
 
     @property
     def is_valid(self):
         jws_verifier = JWSHelper(self.public_jwk)
-        dpop_valid = jws_verifier.verify(self.proof)
+        try:
+            dpop_valid = jws_verifier.verify(self.proof)
+        except Exception as e:
```
peppelinux, line 95: Please specialize this exception,

1. define a new exception in exceptions.py
2. use that exception here

ps = I like the generalization `{e.__class__.__name__}`

peppelinux (changes requested):

pyeudiw/oauth2/dpop.py
```diff
@@ -81,12 +80,24 @@ def __init__(
             if self.dpop_header_prefix in http_header_authz
             else http_header_authz
         )
+        # If the jwt is invalid, this will raise an exception
+        try:
+            unpad_jwt_header(http_header_dpop)
+        except UnicodeDecodeError:
```
peppelinux, line 86: please add a generic Exception after this, in the case that an unhandled exception may occour

do you agree?

salvatorelaiso: Resolved in 560e274

pyeudiw/oauth2/dpop.py
salvatorelaiso: Resolved in 5b531ed

pyeudiw/oauth2/dpop.py
```diff
@@ -5,6 +5,7 @@
 from pydantic import BaseModel, HttpUrl
 from pyeudiw.jwt import JWSHelper
 from pyeudiw.jwt.utils import unpad_jwt_payload, unpad_jwt_header
+from pyeudiw.oauth2.exceptions import KidError
```
peppelinux, line 8: please create/move KidError from oauth2 to the submodule jwk, and import then accordingly here

pyeudiw/oauth2/dpop.py
```diff
@@ -81,12 +81,36 @@ def __init__(
             if self.dpop_header_prefix in http_header_authz
             else http_header_authz
         )
+        # If the jwt is invalid, this will raise an exception
+        try:
+            unpad_jwt_header(http_header_dpop)
+        except UnicodeDecodeError:
+            raise ValueError("DPoP proof is not a valid JWT")
```
peppelinux, line 88: Please add a logger error here

pyeudiw/oauth2/exceptions.py
```diff
@@ -0,0 +1,2 @@
+class KidError(Exception):
```
peppelinux, line 1: please move to jwk since the kid is defined in jwk and not in oauth2

peppelinux (changes requested):

pyeudiw/oauth2/exceptions.py
salvatorelaiso: Resolved in ec43f2e

peppelinux (approved):

PR #36, opened 2023-07-26
feat: [tests] time and datetime

salvatorelaiso:


pyeudiw/tools/utils.py
```diff
@@ -20,8 +19,8 @@ def iat_now() -> int:
 
 
 def exp_from_now(minutes: int = 33) -> int:
-    _now = timezone.localtime()
-    return int((_now + datetime.timedelta(minutes=minutes)).timestamp())
+    now = datetime.datetime.now()
```
peppelinux, line 22: .now is timezone naive

we need a now datetime timezone aware

peppelinux (changes requested):

pyeudiw/tools/utils.py
salvatorelaiso: Resolved in 648d5c0

pyeudiw/tests/tools/test_utils.py
```diff
@@ -0,0 +1,84 @@
...
+import sys
+import freezegun
+
+import pytest
+
+from pyeudiw.tools.utils import exp_from_now, iat_now, random_token
+
+
+def test_make_timezone_aware():
+    # TODO: test the function after it is implemented
```
peppelinux, line 12: please implement that function and test it here

peppelinux (changes requested):

peppelinux (approved):

PR #62, opened 2023-07-31
refactor: from qrcode to pyqrcode

salvatorelaiso:
Switching library for QR code generation.
Inserting a logo on top of an SVG requires further analysis and it shall be discussed.


peppelinux (approved):

pyeudiw/tools/qr_code.py
```diff
@@ -1,65 +1,49 @@
...
-        return stream
-
-    def for_html(self):
-        stream = self.as_svg()
-        data = stream.read().decode()
-        # data = data.replace(
-        # "<?xml version=\'1.0\' encoding=\'UTF-8\'?>\n", ""
-        # )
-        return data
+        qr = pyqrcode.create('http://uca.edu')
```
peppelinux, line 19: 'http://uca.edu' is critical, please align with the data

peppelinux (changes requested):

pyeudiw/tools/qr_code.py
salvatorelaiso: Resolved in eeac00b

PR #71, opened 2023-08-01
feat: add vp token schema

salvatorelaiso:
feat: add vp token schema

### DIFF
--- a/pyeudiw/openid4vp/schema.py
+++ b/pyeudiw/openid4vp/schema.py
@@ -45,6 +45,14 @@ class ResponseSchema(BaseModel):
     presentation_submission: PresentationSubmissionSchema
 
 
+header_model_name = "Header"
+payload_model_name = "Payload"
+cnf_model_name = "Cnf"
+formats_supported_schema = "VpFormatsSupported"
+vp_model_name = "JwtVpJson"
+vc_model_name = "JwtVcJson"
+
+
 class VPToken(BaseModel):
     """
     Schema to validate a VP Token. The token, in the form of a JWS header and payload,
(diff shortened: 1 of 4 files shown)


pyeudiw/openid4vp/schema.py
```diff
@@ -41,3 +43,214 @@ class ResponseSchema(BaseModel):
...
+                         alg=(Literal[
+                             "RS256",
+                             "RS384",
+                             "RS512",
+                             "ES256",
+                             "ES384",
+                             "ES512",
+                             "PS256",
+                             "PS384",
+                             "PS512",
```
peppelinux, line 221: these should not be fixed in the code but taken from a configuration when the backed uses these schema, otherwise these should be taken from a constructor `__init__` if possible or any other strategy but get it our from the sources, these must be parameters (even with a default fallback as workaround)

peppelinux (changes requested):

pyeudiw/openid4vp/schema.py
```diff
@@ -1,8 +1,10 @@
 import re
-from pydantic import BaseModel, ValidationError
-from typing_extensions import Annotated
+from pydantic import BaseModel, ValidationError, create_model, HttpUrl
+from typing_extensions import Annotated, Literal
 from pydantic.functional_validators import AfterValidator
 
+from pyeudiw.jwk.schema import JwkSchema
+
 JWT_REGEX = r"(^[\w-]*.[\w-]*.[\w-]*~([\w-]*.[\w-]*.[\w-]*){1})"
```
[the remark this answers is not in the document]
peppelinux: the SD-JWT schema MUST be moved into the sd_jwt module, do you agree?

pyeudiw/openid4vp/schema.py
```diff
@@ -37,7 +22,225 @@ class PresentationSubmissionSchema(BaseModel):
...
+                Wallet Instance supports the transfer of
+                presentation_definition by
+                reference, with true indicating support.
+    """
+    header: create_model(header_model_name,
+                         alg=(str, ...),
+                         typ=(Literal["wallet-attestation+jwt"], ...),
+                         kid=(str, ...),
+                         x5c=(list[str], ...),
+                         trust_chain=(list[str], ...))
```
peppelinux, line 210: sono opzionali, in assenza di questi si fa discovery

pyeudiw/openid4vp/schema.py
salvatorelaiso: Resolved in [d0d3be0](https://github.com/italia/eudi-wallet-it-python/pull/71/commits/d0d3be073cc01b1cefaad86d5925358b6a982ddf)

pyeudiw/openid4vp/schema.py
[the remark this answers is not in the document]
salvatorelaiso: We need to to do a double check on how the code is interacting. The regex you mentioned was used to validate the field `vp_token` of the response, not the SDJWT itself.

PR #86, opened 2023-08-07
ci: html linting

salvatorelaiso:


salvatorelaiso:
In pause while fixing the HTML files

peppelinux (approved):

PR #96, opened 2023-08-10
Feature: added anchors support

PascalDR:


peppelinux:
the CI fails here
https://github.com/italia/eudi-wallet-it-python/actions/runs/5820025767/job/15779549285?pr=96#step:9:171

the PR presents a lor of copied and pasted code taken from similar methods, a refactor is then required in the future. This increases the costs of this project and is not the way we want to proceed for the next PRs

@PascalDR ^

pyeudiw/storage/db_engine.py
```diff
@@ -142,11 +190,31 @@ def get_trust_attestation(self, entity_id: str) -> Union[dict, None]:
...
+    def get_trust_attestation(self, entity_id: str) -> Union[dict, None]:
+        for db_name, storage in self.storages:
+            try:
+                chain = storage.get_anchor(entity_id)
+
+                if chain:
+                    return chain
+
+            except Exception as e:
+                logger.critical(f"Error {str(e)}")
```
peppelinux, line 203: please merge these two logger calls in a single one

peppelinux (changes requested):
unit tests fails on an http response status code, this is critical for the merge, please fix

PR #101, opened 2023-08-12
v0.5.0

peppelinux:
Consolidations, cleanups, unit tests, trust evaluations, refactors


pyeudiw/sd_jwt/__init__.py
```diff
@@ -47,16 +63,58 @@ def _create_signed_jws(self):
...
+        dp=pk_encode_int(pn.dmp1),
+        dq=pk_encode_int(pn.dmq1),
+        qi=pk_encode_int(pn.iqmp)
+    )
+    return jwcrypto.jwk.JWK(**params)
+
+def _adapt_keys(issuer_key: JWK, holder_key: JWK):
+    # _iss_key = issuer_key.key.serialize(private=True)
+    # _iss_key['key_ops'] = 'sign'
+    _issuer_key = import_pyca_pri_rsa(issuer_key.key.priv_key, kid=issuer_key.kid)
```
peppelinux, line 109: this works just with RSA and not with EC

we must support both

Issue #111, opened 2023-08-23
Features by Metadata


peppelinux:
in the current implementation (0.6.0) the metadata is published within the entity configuration in an arbitrary way, while the metadata protocol specific features, requirements and capabilities defined within it are not implemented yet.

We need to implement the followings:

- [x] `default_acr_values`: checks that the LoA during the presentation satisfies the supported values
- [x] `default_max_age`: defines how long an authenticated session, pertaining a presentation, is considered as valid. After this time period the authenticated session must expire. Specifies that the End-User MUST be actively authenticated if the End-User was authenticated longer ago than the specified number of seconds. 
- [x] `id_token_ ...`, related to SIOPv2. they we'll be implemented when the pseudonyms will be implemented
- [x] `presentation_definitions`, the RP Must check that the disclosed attributes within the presentation matches the one required as definite in the presentation definition
- [x] `require_auth_time`, boolean value specifying whether the `auth_time` Claim in the ID Token is REQUIRED. Related to SIOPv2.
- [x] `subject_type`, already implemented. An additional qualitative check is required.
- [x] `vp_formats`, check that the presentation format is compliant to the values defined in this parameter.

peppelinux:
We need two different kind of validation:

- schema based, according to this example: https://github.com/italia/spid-cie-oidc-django/tree/main/spid_cie_oidc/entity/schemas
- feature based, according to the metadata member that actually are implemented. Here we may issu a warning during the first execution of the proxy with the list of all the unsupported features

PascalDR:
Should be resolved in PR [138](https://github.com/italia/eudi-wallet-it-python/pull/138).

PR #138, opened 2023-10-27
Feat/features by metadata

PascalDR:
Relative to issue [https://github.com/italia/eudi-wallet-it-python/issues/111](111).

This PR adds initial support to metadata validation adding a warning print at start of execution.


pyeudiw/federation/schemas/wallet_relying_party.py
```diff
@@ -1,45 +1,51 @@
-from typing import Any, Dict, List
+from enum import Enum
```
peppelinux, line 1: we need to distiguish th entity configuration in federation and the RP metadata in openid4vp

all about openid4vp must be movedi n the openid4vp package and the EC in federation must use the metadata defined there

this allow us to use rp metadata even outside federation

peppelinux (changes requested):

pyeudiw/federation/schemas/wallet_relying_party.py
PascalDR: I moved the the code relative to VP to the relative package in commit 98d494d.
It's ok?

peppelinux: wallet_relying_party is a federation name

please, use a name agnostic to the trust model, wallet_relying_party.py -> __init__.py if it is related to openid4vp.schemas

pyeudiw/satosa/backend.py
```diff
@@ -33,8 +33,10 @@
 )
 from pyeudiw.storage.db_engine import DBEngine
 from pyeudiw.storage.exceptions import StorageWriteError
+from pyeudiw.federation.schemas.wallet_relying_party import WalletRelyingParty
 
 from pydantic import ValidationError
+from pprint import pformat
```
peppelinux, line 39: if pformat uses multiplelines, It cannot be used for logs because it may run in conflicts with the syslog approached and the common logging systems

then: do not USE multiple .warnings, please use format strings in a single warning

PascalDR: Resolved in [4165826](https://github.com/italia/eudi-wallet-it-python/pull/138/commits/4165826cec281b07f5f4e0912984b240e1170ebd) and [364331e](https://github.com/italia/eudi-wallet-it-python/pull/138/commits/364331ed8cf82e0594b6369383d977c670a0f9a2)

pyeudiw/federation/schemas/wallet_relying_party.py
PascalDR: Resolved in [33d8000](https://github.com/italia/eudi-wallet-it-python/pull/138/commits/33d80005e1d86e8f613dc60f979c7af127aa9d1c)

Issue #170, opened 2023-11-17
`./jwk/schema.py:9`:    # TODO: verify if is optional


davidelongo-ey:

salvatorelaiso:
This TODO refers to the following:

https://github.com/italia/eudi-wallet-it-python/blob/43454bf31d22f67b0c5ee6a66da69efe1c27f9f0/pyeudiw/jwk/schema.py#L9-L22


According to the [RFC 7517 section 4.4](https://datatracker.ietf.org/doc/html/rfc7517#section-4.4) it is indeed optional.
Can we close this issue by simply removing its TODO in a dedicated PR or do we prefer to enforce the use of `alg` nonetheless?

peppelinux:
@salvatorelaiso got It

Please remove that todo by reporting in the PR the evidence taken from rfc7517 and then close this issue

Thx

PR #187, opened 2023-11-22
chore: remove TODO

salvatorelaiso:
According to [RFC7517 section 4.4](https://datatracker.ietf.org/doc/html/rfc7517#section-4.4), `alg` is optional .


salvatorelaiso:
Closes #170

peppelinux (approved):

Issue #193, opened 2023-11-29
Presentation Definition


peppelinux:
we have to implement this schema
https://vcstuff.github.io/oid4vc-haip-sd-jwt-vc/draft-oid4vc-haip-sd-jwt-vc.html#appendix-B

PR #196, opened 2023-12-05
feat: presentation definition schema  Pydantic model

salvatorelaiso:
Generated schema with `datamodel-codegen` tool.

The resulting schema is updated to reflect changes in Pydantic V2.0

The model has been generated starting from the schema [Presentation Definition](https://openid.github.io/oid4vc-haip-sd-jwt-vc/draft-oid4vc-haip-sd-jwt-vc.html#appendix-B) ([GitHub](https://github.com/openid/oid4vc-haip-sd-jwt-vc/blob/main/schemas/presentation_definition.json)).

This json-schema presents a reference to [Claim format designations](http://identity.foundation/claim-format-registry/schemas/presentation-definition-claim-format-designations.json) to specify the values allowed in `input descriptor`'s `format` field ([ref](
https://github.com/openid/oid4vc-haip-sd-jwt-vc/blob/75a17220959a28a208b04546d4337b9b5ed5c5bd/schemas/presentation_definition.json#L49-L50)).

The values must be matched by one of the following regex:
- `"^jwt$|^jwt_vc$|^jwt_vp$"`
- `"^ldp_vc$|^ldp_vp$|^ldp$"`

None of the above matches the value `vc+sd-jwt`, which is the expected value according to https://openid.github.io/oid4vc-haip-sd-jwt-vc/draft-oid4vc-haip-sd-jwt-vc.html#name-presentation-definition.

@peppelinux What do we do? We can either extend the above matching by including a case which captures the value `vc+sd-jwt` or simply drop the regular expressions and match **exactly** this value.


peppelinux:
please add in the documentation in the repository how you have used datamodel-codegen for producing the pydantic schema, with an example

this will help contributors in aligning the code when future release of the sd-jwt-vc schema will be provided in the official specs

peppelinux:
> What do we do? We can either extend the above matching by including a case which captures the value vc+sd-jwt or simply drop the regular expressions and match exactly this value.

it depends by the format supported, this behaviour is linked to this
https://github.com/italia/eudi-wallet-it-python/issues/111

for the current milestone we just have to support sd-jwt, the next will have mdoc cbor as well

salvatorelaiso:
Closes #193

PR #200, opened 2023-12-11
Elliptic curve support

### LINKED ISSUE #102: Tests for Elliptic Curve
The sd-jwt issuance is tested with RSA and not with EC
we must support both

_Originally posted by @peppelinux in https://github.com/italia/eudi-wallet-it-python/pull/101#discussion_r1295256503_

salvatorelaiso:
This PR extends JWEHelper class capabilities to handle encryption and decryption with the use of EC Keys.
The above functionalities are tested in the corresponding unit tests.

It also adds EC keys based integration test with new files while leaving the RSA integration test intact.

This PR provides a new JWK Schema as per TODO: https://github.com/italia/eudi-wallet-it-python/blob/78c00cba554bd789f70bb98e95c97a29c8fb5dae/pyeudiw/oauth2/dpop/schema.py#L20-L21
The usage of this dynamic schema handling by Pydantic is documented through a simple test added in the file `test_jwk`, named `test_dynamic_schema_validation`.

The old schemas are retained for compatibility purpose until migration to this newer schemas.

Closed #102


example/satosa/integration_test/main_ec.py
```diff
@@ -0,0 +1,283 @@
+import requests
```
peppelinux, line 1: please do not duplicate similar code, replace the main with this one where EC is used by default or create an argparse prompt where options can be used to defined if using RSA or EC or which alg to use

example/satosa/integration_test/settings_ec.py
```diff
@@ -0,0 +1,135 @@
...
+    EXP,
+    leaf_wallet_jwk,
+    ta_ec,
+    ta_jwk
+)
+
+from pyeudiw.tools.utils import iat_now, exp_from_now
+
+
+RP_EID = "https://localhost:10000/OpenID4VP"
```
peppelinux, line 15: localhost:10000 not exists anymore, https://localhost is now our demo FQDN

pyeudiw/sd_jwt/__init__.py
```diff
@@ -105,12 +105,29 @@ def import_pyca_pri_rsa(key, **params):
     )
     return jwcrypto.jwk.JWK(**params)
 
+def import_ec(key, **params):
+    pn = key.private_numbers()
+    params.update(
+        kty="EC",
+        crv="P-256",
```
peppelinux, line 112: don't use constants but configurable parameters

here we're forcing P-256 within the code, while we must take it from the args and we may have a default fallback value

peppelinux (changes requested):
Small things to be fixed before the merge

peppelinux (approved):

PR #216, opened 2024-01-02
chore: some additional logs

peppelinux:
proofread of the current project with a particular attention to logging and exception handling

### DIFF
--- a/pyeudiw/federation/trust_chain_validator.py
+++ b/pyeudiw/federation/trust_chain_validator.py
@@ -132,18 +132,27 @@ class StaticTrustChainValidator:
         )
 
         if not ta_jwk:
+            logger.error(
+                f"Trust chain validation error: TA jwks not found."
+            )
             return False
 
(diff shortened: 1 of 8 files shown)


pyeudiw/federation/trust_chain_validator.py
```diff
@@ -204,16 +220,16 @@ def _retrieve_es(self, download_url: str, iss: str) -> str:
         """
         jwt = get_entity_statements(download_url, self.httpc_params)
         if not jwt:
-            logger.warning(
-                f"Cannot fast refresh Entity Statement {iss}"
-            )
+            _msg = f"Cannot fast refresh Entity Statement {iss}"
+            logger.warning(_msg)
+            #  raise HttpError(_msg)
```
peppelinux, line 225: we should normalize this behaviour by enabling the raise of the exception.

at this moment I have disabled this due to the current not handling of this exception when the trust chain is fastly updated while some nodes are not available (that might happen in the real world)

Issue #33, opened 2023-07-25
Retention rule Session Storage


Pyscho79:
Create to configure retention rule session

- [x] Create Doc
- [x] Create Configuration param (default: 2 years)
- [x] Add parameter in the configuration (storage.init_params.conf.data_ttl: in seconds)
- [x] Integrate TTL rule index in the constructor

Following this documentation: https://www.mongodb.com/docs/manual/tutorial/expire-data/

peppelinux:
Done https://github.com/italia/eudi-wallet-it-python/pull/218

PR #218, opened 2024-01-08
[Feat/retention rule] Added ttl rule for sessions

PascalDR:
This PR is relative to issue [33](https://github.com/italia/eudi-wallet-it-python/issues/33) and adds the ttl rule to the session collection on mongo storage class.


pyeudiw/storage/mongo_storage.py
```diff
@@ -124,6 +127,19 @@ def init_session(self, document_id: str, session_id: str, state: str) -> str:
         self.sessions.insert_one(entity)
 
         return document_id
+    
+    def set_session_retention_ttl(self, ttl: int | None) -> None:
+        self._connect()
+
+        if ttl == None:
```
peppelinux, line 134: None can be expressed as a non existent or False, and we can have it as an int valued to 0 (zero) as well

peppelinux, line 134:
```suggestion
        if not ttl:
```

pyeudiw/storage/mongo_storage.py
```diff
@@ -124,6 +127,19 @@ def init_session(self, document_id: str, session_id: str, state: str) -> str:
         self.sessions.insert_one(entity)
 
         return document_id
+    
+    def set_session_retention_ttl(self, ttl: int | None) -> None:
```
peppelinux, line 131:
```suggestion
    def set_session_retention_ttl(self, ttl: int) -> None:
```

peppelinux (changes requested):
I made some quick changes in the form of suggestion to be merged by you @PascalDR 

at the same time I expect to have in this PR also the settings parameter in the storage section where the ttl value can be configured as well

pyeudiw/storage/base_storage.py
```diff
@@ -40,6 +40,24 @@ def init_session(self, document_id: str, dpop_proof: dict, attestation: dict) ->
         :param attestation: the attestation.
         """
         raise NotImplementedError()
+    
+    def set_session_retention_ttl(self, ttl: int | None) -> None:
```
peppelinux, line 44:
```suggestion
    def set_session_retention_ttl(self, ttl: int) -> None:
```

peppelinux (changes requested):

peppelinux:
Si @PascalDR te l'ho esplicitamente chiesto

> at the same time I expect to have in this PR also the settings parameter in the storage section where the ttl value can be configured as well

peppelinux (approved):

PR #235, opened 2024-01-23
fix: introduce constraint on cryptography version

salvatorelaiso:
Introduce an upperbound constraint on the package `cryptography` to resolve the problems caused by the latest release `42.0.0`.

Additional info on the breaking changes and the errors can be found [here](https://github.com/salvatorelaiso/eudi-wallet-it-python/actions/runs/7630085474/job/20784982336).

Closes #234

PR #236, opened 2024-01-23
fix: deps, cryptography breaking changes

peppelinux:
this PR resolves the breaking change found in the last version of the python package cryptography.
It exceeds and closes https://github.com/italia/eudi-wallet-it-python/pull/235


salvatorelaiso (commented):
LGTM

PR #237, opened 2024-01-25
Docs storage

### LINKED ISSUE #232: Storage Documentation
This file was not aligned with the updates

https://github.com/italia/eudi-wallet-it-python/blob/dev/docs/STORAGE.md

and more: it still needs explanatory text.

salvatorelaiso:
Update `STORAGE.md` file with references and a briefly introduction to the classes structure.

Closes #232

### DIFF
--- a/docs/STORAGE.md
+++ b/docs/STORAGE.md
@@ -67,51 +67,69 @@ This classes can be used as references while providing a custom implementation f
 
 #### Trust Anchors
 
-````
-    "trust_anchors": {
-       "https://ta.example.org": {
-          "federation": {
-            "entity_configuration": "str(EC), -> EC contains the federation entity public keys,
-            "exp": datetime
-         },
-          "x509": {
-            "pem": str(PEM) -> contains public keys,
-            "exp": datetime
-       }
+```json
+[  
+  {
+    "_id": ObjectId,
+    "entity_id": string,
+    "federation": {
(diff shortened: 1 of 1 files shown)


docs/STORAGE.md
```diff
@@ -1,18 +1,69 @@
...
+In the realm of pyeudiw, seamless integration with MongoDB is facilitated through specialized classes, namely 
+[`MongoStorage`](../pyeudiw/storage/mongo_storage.py) and [`MongoCache`](../pyeudiw/storage/mongo_cache.py).  
 
+These classes not only offer a robust implementation but also serve as tangible representations of the abstract base 
+classes, `BaseStorage` and `BaseCache`.
+This classes can be used as references while providing a custom implementation for other databases.
+
+## Data Examples
 
 ### Trust Anchors
```
peppelinux, line 69: @salvatorelaiso do you confirm that the schema is not changed?

could you please explain the content and the scope of each element of this json object?

peppelinux (changes requested):

docs/STORAGE.md
salvatorelaiso: What are you referring to with "schema"? The schema of the classes and their relationships?
If yes, then the answer is affirmative.
Will add more details on the json object.

peppelinux: please add more details regarding the json object and also say that's only related to the mongodb schema

peppelinux (approved):

PR #257, opened 2024-08-30
feat: response_code in redirect_uri

### LINKED ISSUE #251: [Satosa] Authorization Response and after-authentication flow is outdated
The provided authorization resposnse handler has an outdated behaviour in regard to the protocol flow after the authorization respose is sent by the Wallet to the Verifier.

This is the current response to the POST Authorization Request (executed by the Wallet toward the Verifier).
https://github.com/italia/eudi-wallet-it-python/blob/52fcbece3fa77fefbd5e91181ac3e2e73fc5f098/pyeudiw/satosa/default/request_handler.py#L191-L203
Here the redirect URI (when present) is a fixed URI but it should intead include a fresh, random and cyrptographically secure code as described in [IT Wallet specifications](https://italia.github.io/eudi-wallet-it-docs/versione-corrente/en/relying-party-solution.html#:~:text=The%20redirect_uri%20value,or%20Cross%20device.) and [OID4VP Draft21](https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#name-response-mode-direct_post:~:text=The%20value%20of%20the%20redirect,and%20process%20the%20Authorization%20Response).

A smilar issue arise in the cross device flow, when the redirect URI is not provided as a response ot the Auth Request but is included insteead in the as a response in the `/status` request. This is the relevant code
https://github.com/italia/eudi-wallet-it-python/blob/52fcbece3fa77fefbd5e91181ac3e2e73fc5f098/pyeudiw/satosa/default/openid4vp_backend.py#L306-L311

Again, here the query parameter `id={state}` in the redirect uri should be replaced with a fresh code.

In either flow, the after-authentication endpoint/callback should accept and process the new code. In the current version, it processes the id instead
https://github.com/italia/eudi-wallet-it-python/blob/52fcbece3fa77fefbd5e91181ac3e2e73fc5f098/pyeudiw/satosa/default/openid4vp_backend.py#L226-L231
Note that current version is broken as the same device flow does not include parameter `id={state}`; this always fails in the integration tests.

How this new fresh code should be binded by the response handler to the user data provided by the Wallet is up to the implementation. OID4VP suggests an [implementation flow](https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-wg-draft.html#name-response-mode-direct_post-2) but to me the exact details on how to port this approach on the Satosa backend are unclear.

Tagging @peppelinux as this issue is non-trivial and I am not sure where this fit across the current milestones.

Zicchio:
This pull request advances the post authentication flow to what is described in version 0.8.0 of https://github.com/italia/eudi-wallet-it-docs
It introduces a redirect uri containing a `response_code` that is opaque to the Wallet Instance and/or User Agent (based on cross device or same device flow).
This closes #251 by implementing strategy 3 discussed here https://github.com/italia/eudi-wallet-it-python/issues/251#issuecomment-2320502845
AES-GCM was used as a the algorithmic choice of authenticated encryption as suggested by the book *Real-World Cryptography* by David Wong, chapter 4.5 .

### DIFF
--- a/pyeudiw/satosa/default/request_handler.py
+++ b/pyeudiw/satosa/default/request_handler.py
@@ -16,7 +16,6 @@ class RequestHandler(RequestHandlerInterface, BackendDPoP, BackendTrust):
     def request_endpoint(self, context: Context, *args) -> JsonResponse:
         self._log_function_debug("response_endpoint", context, "args", args)
 
-        # TODO: funzionalità per metodo POST
 
         try:
             state = context.qs_params["id"]
--- a/pyeudiw/satosa/default/request_handler.py
+++ b/pyeudiw/satosa/default/request_handler.py
@@ -16,6 +16,8 @@ class RequestHandler(RequestHandlerInterface, BackendDPoP, BackendTrust):
     def request_endpoint(self, context: Context, *args) -> JsonResponse:
         self._log_function_debug("response_endpoint", context, "args", args)
 
+        # TODO: funzionalità per metodo POST
+
         try:
             state = context.qs_params["id"]
         except Exception as e:
@@ -38,7 +40,6 @@ class RequestHandler(RequestHandlerInterface, BackendDPoP, BackendTrust):
             "iat": iat_now(),
             "exp": exp_from_now(minutes=self.config['authorization']['expiration_time'])
         }
-
         # take the session created in the pre-request authz endpoint
         try:
             document = self.db_engine.get_by_state(state)
--- a/pyeudiw/satosa/utils/respcode.py
+++ b/pyeudiw/satosa/utils/respcode.py
@@ -1,0 +1,18 @@
+import base64
+import hashlib
+import hmac
+
+
+def create_response_code(state: str, key: str) -> str:
+    bkey = base64.b64decode(key)
+    code = hmac.new(bkey, msg=state.encode(), digestmod=hashlib.sha256)
+    b64code = base64.urlsafe_b64encode(code.digest()).decode().strip("=")
+    return b64code
+
+
+def validate_resp_code(obt_b64code: str, state: str, key: str) -> bool:
+    bkey = base64.b64decode(key)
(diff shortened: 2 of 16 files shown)


pyeudiw/satosa/default/request_handler.py
```diff
@@ -16,6 +16,8 @@ class RequestHandler(RequestHandlerInterface, BackendDPoP, BackendTrust):
     def request_endpoint(self, context: Context, *args) -> JsonResponse:
         self._log_function_debug("response_endpoint", context, "args", args)
 
+        # TODO: funzionalità per metodo POST
```
peppelinux, line 19:
```suggestion
```

pyeudiw/satosa/utils/respcode.py
```diff
@@ -0,0 +1,80 @@
+import base64
+from dataclasses import dataclass, field
+from cryptography.hazmat.primitives.ciphers.aead import AESGCM
+import secrets
+
```
peppelinux, line 5:
```suggestion
import string

```

pyeudiw/satosa/utils/respcode.py
```diff
@@ -0,0 +1,80 @@
...
+CODE_SYM_KEY_LEN = 32  # in bytes (256 bits)
+_HEX_ALPHABETH = '1234567890abcdef'
```
peppelinux, line 7:
```suggestion
```

pyeudiw/satosa/utils/respcode.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def create_code(self, state: str) -> str:
+        return create_code(state, self.key)
+
+    def recover_state(self, code: str) -> str:
+        return recover_state(code, self.key)
+
+
+def decode_key(key: str) -> bytes:
+    if not set(key) <= set(_HEX_ALPHABETH):
```
peppelinux, line 39:
```suggestion
    if not set(key) <= set(string.hexdigits):
```

peppelinux (approved):

PR #258, opened 2024-09-02
Feat: check for previous authn session

### LINKED ISSUE #109: pre auth endpoint must check the session state and the previous authn session
we need to filter out the requests that lands on the disco page without having processed through a SAML2/OIDC frontend

this is required for both UX and security

elisanp:
Closes #109 by filtering out requests that have not been processed through any frontend

### DIFF
--- a/pyeudiw/satosa/default/openid4vp_backend.py
+++ b/pyeudiw/satosa/default/openid4vp_backend.py
@@ -176,10 +176,10 @@ class OpenID4VPBackend(OpenID4VPBackendInterface, BackendTrust):
         session_id = context.state["SESSION_ID"]
         state = str(uuid.uuid4())
 
-        self._log_warning(
-            context,
-            "Preventing session creation if context is not linked to any previous authn session not implemented yet"
-        )
+        if not context.target_frontend:
+            _msg = f"Preventing session creation. Context is not linked to any previous authn session."
+            self._log_info(context, _msg)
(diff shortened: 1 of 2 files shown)


pyeudiw/satosa/default/openid4vp_backend.py
```diff
@@ -176,10 +176,10 @@ def pre_request_endpoint(self, context: Context, internal_request, **kwargs) ->
         session_id = context.state["SESSION_ID"]
         state = str(uuid.uuid4())
 
-        self._log_warning(
-            context,
-            "Preventing session creation if context is not linked to any previous authn session not implemented yet"
-        )
+        if not context.target_frontend:
+            _msg = "Preventing session creation: context is not linked to any previous authn session."
+            self._log_info(context, _msg)
```
peppelinux, line 181:
```suggestion
            self._log_warn(context, _msg)
```

pyeudiw/satosa/default/openid4vp_backend.py
```diff
@@ -176,10 +176,10 @@ def pre_request_endpoint(self, context: Context, internal_request, **kwargs) ->
...
+            return self._handle_400(context, "valid previous authn session not found")
```
peppelinux, line 182:
```suggestion
            return self._handle_400(context, "previous authn session not found. It seems that the flow did not started with a valid authn request to one of the configured frontend.")
```

PR #304, opened 2024-11-11
fix: qr code value

### LINKED ISSUE #302: [Satosa][Bug] Incorrect QR code value in cross device flow
The cross device flow containg a QR code with the wrong value: the link host is not satosa-pyeudiw itself but it should be the target wallet application.
This is the line of code with the incorrect behaviour
https://github.com/italia/eudi-wallet-it-python/blob/6a658f27d55a04129ca21c72e535f84d3b255004/pyeudiw/satosa/default/openid4vp_backend.py#L215-L218

The QR code should be a deep-link/universal-link/whatever much like in the same device flow as otherwise the user would not be able to open the holder app.
In fact, it's the same link.

See
https://italia.github.io/eudi-wallet-it-docs/versione-corrente/en/relying-party-solution.html#authorization-request-details

Zicchio:
This pull requests closes #302 

The cross device flow now uses the same URL of the same device flow, and the URL value is the very same value that was previously yielded by the same device flow.

### DIFF
--- a/pyeudiw/satosa/default/openid4vp_backend.py
+++ b/pyeudiw/satosa/default/openid4vp_backend.py
@@ -344,6 +344,8 @@ class OpenID4VPBackend(OpenID4VPBackendInterface, BackendTrust):
         scheme = self.config["authorization"]["url_scheme"]
         if "://" not in scheme:
             scheme = scheme + "://"
+        if not scheme.endswith("/"):
+            scheme = scheme + "/"
         # NOTE: path component is currently unused by the protocol, but currently
         # we leave it there as 'authorize' to stress the fact that this is an
         # OAuth 2.0 request modified by JAR (RFC9101)
--- a/pyeudiw/satosa/default/openid4vp_backend.py
+++ b/pyeudiw/satosa/default/openid4vp_backend.py
@@ -205,20 +205,18 @@ class OpenID4VPBackend(OpenID4VPBackendInterface, BackendTrust):
             'client_id': self.client_id,
             'request_uri': f"{self.absolute_request_url}?id={state}",
         }
-        url_params = urlencode(payload, quote_via=quote_plus)
+
+        respose_url = self._build_authz_request_url(payload)
 
         if is_smartphone(context.http_headers.get('HTTP_USER_AGENT')):
             # Same Device flow
-            res_url = f'{self.config["authorization"]["url_scheme"]}://authorize?{url_params}'
-            return Redirect(res_url)
+            return Redirect(respose_url)
 
         # Cross Device flow
-        res_url = f'{self.client_id}?{url_params}'
-
         result = self.template.qrcode_page.render(
             {
                 "qrcode_color": self.config["qrcode"]["color"],
-                "qrcode_text": res_url,
+                "qrcode_text": respose_url,
                 "qrcode_size": self.config["qrcode"]["size"],
                 "qrcode_logo_path": self.config["qrcode"]["logo_path"],
                 "qrcode_expiration_time": self.config["qrcode"]["expiration_time"],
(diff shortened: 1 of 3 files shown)


pyeudiw/satosa/default/openid4vp_backend.py
```diff
@@ -205,20 +205,18 @@ def pre_request_endpoint(self, context: Context, internal_request, **kwargs) ->
             'client_id': self.client_id,
             'request_uri': f"{self.absolute_request_url}?id={state}",
         }
-        url_params = urlencode(payload, quote_via=quote_plus)
+
+        respose_url = self._build_authz_request_url(payload)
```
peppelinux, line 209:
```suggestion
        response_url = self._build_authz_request_url(payload)
```

pyeudiw/satosa/default/openid4vp_backend.py
```diff
@@ -205,20 +205,18 @@ def pre_request_endpoint(self, context: Context, internal_request, **kwargs) ->
...
 
         if is_smartphone(context.http_headers.get('HTTP_USER_AGENT')):
             # Same Device flow
-            res_url = f'{self.config["authorization"]["url_scheme"]}://authorize?{url_params}'
-            return Redirect(res_url)
+            return Redirect(respose_url)
```
peppelinux, line 213:
```suggestion
            return Redirect(response_url)
```

pyeudiw/satosa/default/openid4vp_backend.py
```diff
@@ -205,20 +205,18 @@ def pre_request_endpoint(self, context: Context, internal_request, **kwargs) ->
...
 
         # Cross Device flow
-        res_url = f'{self.client_id}?{url_params}'
-
         result = self.template.qrcode_page.render(
             {
                 "qrcode_color": self.config["qrcode"]["color"],
-                "qrcode_text": res_url,
+                "qrcode_text": respose_url,
```
peppelinux, line 219:
```suggestion
                "qrcode_text": response_url,
```

pyeudiw/satosa/default/openid4vp_backend.py
```diff
@@ -342,6 +340,15 @@ def db_engine(self) -> DBEngine:
...
         return self._db_engine
 
+    def _build_authz_request_url(self, payload: dict) -> str:
+        scheme = self.config["authorization"]["url_scheme"]
+        # NOTE: path component is currently unused by the protocol, but currently
+        # we leave it there as 'authorize' to stress the fact that this is an
+        # OAuth 2.0 request modified by JAR (RFC9101)
+        path = "authorize"
+        query_params = urlencode(payload, quote_via=quote_plus)
+        return f"{scheme}://{path}?{query_params}"
```
peppelinux, line 350:
```suggestion
        return f"{scheme}://{path}?{query_params}"
```

this would work only with custom url schemes, using universal links "://" will be duplicated and wrongly appended

I would suggest something like this

```suggestion
        if "://" in scheme:
            scheme = scheme
        else:
           scheme = f"{scheme}://"
        return f"{scheme}{path}?{query_params}"
```

Zicchio: to support univesal link, further changes were added; this includes

1. added to docs that the same configuration can be used for costum scheme or universal link
2. a check for the existance of trailing slash before path component is added (required if universal link does not include them)
3. an extension of same device integration test to make it compatible with universal links, as in that case a different exception is raised

pyeudiw/satosa/default/openid4vp_backend.py
```diff
@@ -342,6 +340,19 @@ def db_engine(self) -> DBEngine:
 
         return self._db_engine
 
+    def _build_authz_request_url(self, payload: dict) -> str:
+        scheme = self.config["authorization"]["url_scheme"]
+        if "://" not in scheme:
+            scheme = scheme + "://"
+        if not scheme.endswith("/"):
+            scheme = scheme + "/"
```
peppelinux, line 348: not sure about this

haip:///?

while haip:// would be good

there is no requirements in having https urls with pending slashes, like https://example.org/

Issue #299, opened 2024-11-05
Define Presentation Submission Requirements for Verifiable Presentations


LadyCodesItBetter:
Implement “presentation_submission” to evaluate the presentation sent by the wallet. 

This module MUST be included within the package openid4vp and defines which attributes are necessary and the conditions to validate credentials within a presentation.

LadyCodesItBetter:
PR #314

peppelinux:
@Zicchio since we have brought back presentation definition for potential LSP, it makes sense to bring back also the presentation submission to be integrated in the response handler

PR #314, opened 2024-11-22
#299-feature/presentation submission dynamic handler

LadyCodesItBetter:
This PR introduces enhancements to the PresentationSubmission class to improve error handling and dynamic handler initialization based on descriptor_map.

Key points:
 - handlers are dynamically created for each descriptor_map entry based on the format key.
 - handlers attribute contains a dictionary where the key is the index of the descriptor, and the value is the respective handler instance.


pyeudiw/openid4vp/presentation_submission/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+        Returns:
+            Dict[int, object]: A dictionary mapping indices to handler instances.
+
+        Raises:
+            KeyError: If the 'format' key is missing in any descriptor.
+            ValueError: If a format is not supported or not defined in the configuration.
+            ImportError: If a module or class cannot be loaded.
+        """
+        handlers = {}
+
```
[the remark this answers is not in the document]
peppelinux: please put some limits on the presentation submission size, in a way that an attacker might not push 10mb of presentation submission causing resource exaustion to the RP

pyeudiw/openid4vp/presentation_submission/__init__.py
```diff
@@ -0,0 +1,87 @@
+import os
+import yaml
+import importlib
+from typing import Dict, Any
+import logging
+
+logger = logging.getLogger(__name__)
+
+class PresentationSubmission:
```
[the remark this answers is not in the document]
LadyCodesItBetter: Done

pyeudiw/openid4vp/presentation_submission/__init__.py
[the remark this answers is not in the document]
LadyCodesItBetter: I've added control for byte size and credential number too

pyeudiw/openid4vp/presentation_submission/config.yml
```diff
@@ -0,0 +1,15 @@
...
+    module: "pyeudiw.openid4vp.presentation_submission"
+    class: "LdpVp"
+  - name: "jwt_vp_json"
+    module: "pyeudiw.openid4vp.presentation_submission"
+    class: "JwtVpJson"
+  - name: "ac_vp"
+    module: "pyeudiw.openid4vp.presentation_submission"
+    class: "AcVp"
+
+MAX_SUBMISSION_SIZE: 10 * 1024 * 1024
```
LadyCodesItBetter, line 15:
```suggestion
MAX_SUBMISSION_SIZE: 4096 * 100
```

pyeudiw/openid4vp/presentation_submission/config.yml
```diff
@@ -0,0 +1,15 @@
...
+    module: "pyeudiw.openid4vp.presentation_submission"
+    class: "LdpVp"
+  - name: "jwt_vp_json"
+    module: "pyeudiw.openid4vp.presentation_submission"
+    class: "JwtVpJson"
+  - name: "ac_vp"
+    module: "pyeudiw.openid4vp.presentation_submission"
+    class: "AcVp"
+
+MAX_SUBMISSION_SIZE: 4096 * 100
```
LadyCodesItBetter, line 15:
```suggestion
MAX_SUBMISSION_SIZE: 4096
```

pyeudiw/openid4vp/presentation_submission/presentation_submission.py
```diff
@@ -0,0 +1,122 @@
+import os
+from pydantic import ValidationError
+import yaml
+import importlib
+from typing import Dict, Any
```
Zicchio, line 5: Importing `Dict` is deprecated since Python 3.9 and this project officially supports Python 3.10+ only - use `dict` instead (no import required).
https://docs.python.org/3/library/typing.html#aliases-to-built-in-types

pyeudiw/openid4vp/presentation_submission/presentation_submission.py
```diff
@@ -0,0 +1,122 @@
...
+        """
+        Load the configuration from format_config.yml located in the same directory.
+
+        Returns:
+            Dict[str, Any]: The configuration dictionary.
+
+        Raises:
+            FileNotFoundError: If the configuration file is not found.
+        """
+        config_path = os.path.join(os.path.dirname(__file__), "config.yml")
```
Zicchio, line 40: What's the point of having a class configurable by editing an obscure configuration file in the *installed* class project that is available who-knows-where based on what `pip` (or other build tool) is used?

@peppelinux @LadyCodesItBetter I'm not sure here what is the intended design goal.

Zicchio:
>handlers attribute contains a dictionary where the key is the index of the descriptor, and the value is the respective handler instance.

Sorry if the question is dull, but... what is an handler?
Let's say I create an instance of `PresentationSubmission`. How and when am I supposed to use `self.handler`? What should they do?

pyeudiw/openid4vp/presentation_submission/presentation_submission.py
```diff
@@ -0,0 +1,122 @@
...
+                raise ValueError(f"Format '{format_name}' is not supported or not defined in the configuration.")
+
+            module_name = format_conf["module"]
+            class_name = format_conf["class"]
+
+            try:
+                # Dynamically load the module and class
+                module = importlib.import_module(module_name)
+                cls = getattr(module, class_name)
+                handlers[index] = cls()  # Instantiate the class
```
Zicchio, line 114: AFAIK the fact that each handler is a class initialized with an empty constructor implies that handlers are functionally just namespace for static methods. It's impossible to known what they should do without doing introspection on what aforementioned namespace contains.

pyeudiw/tests/openid4vp/presentation_submission.py
```diff
@@ -0,0 +1,121 @@
+import pytest
```
Zicchio, line 1: maybe I'm wrong but if a test class name does not start with `test_` or ends with `_test.py` then `pytest` will not execute them for unit tests (altough ci-cd pipeline might hange this behaviour - I'm not 100% sure how unit tests are configured here).
https://docs.pytest.org/en/stable/explanation/goodpractices.html#test-discovery

pyeudiw/openid4vp/presentation_submission/schemas.py
```diff
@@ -0,0 +1,23 @@
...
+    id: str
+    format: str
+    path: str
+    path_nested: Dict[str, Any] = None
+
+
+class PresentationSubmissionSchema(BaseModel):
+    id: str
+    definition_id: str
+    descriptor_map: List[DescriptorSchema]
```
Zicchio, line 15: `List` is also deprecated: use `list` instead

pyeudiw/openid4vp/presentation_submission/presentation_submission.py
peppelinux: the intended propose is for a generic python package, not necessarly used in the iam proxy context

pyeudiw/tests/openid4vp/presentation_submission.py
peppelinux: I can confirm that _test doesn't get loaded

pyeudiw/openid4vp/presentation_submission/presentation_submission.py
Zicchio: The point still stands. Assume I am using the project as a package, that is, I am using it with `pip install pyeudiw` as a part of my project.
To change the configuration of this class, I have to go in the location where pip placed `config.yaml` (which is usually inside the `site-packages`) and and edit it. This is IMO very convoluted and only doable by technically advanced users.

peppelinux: yeah, absolutely. We cannot rely on the relative path of tha file distributed within a python package.

LadyCodesItBetter: To address the issue of accessibility and management of the `config.yml` file, I propose two possible solutions to enhance flexibility and simplify the use of the library:
- dynamic configurations via env of configuration file (a mechanism to dynamically specify the configuration file path using an environment variable, such as `PYEUDIW_CONFIG_PATH`)
- inline configuration support (allow configuration to be passed directly as a Python dictionary or JSON object during class initialization)

@peppelinux which do you think would work better for our use case?

pyeudiw/openid4vp/presentation_submission/presentation_submission.py
LadyCodesItBetter: The PresentationSubmission class processes and validates presentation submissions. Its handlers attribute is a dictionary mapping indices from the descriptor_map in the submission to dynamically instantiated handler objects. These handlers are responsible for specific operations (e.g., validation or transformation) tied to the format defined in a configuration file (config.yml).

The handlers are not just namespaces for static methods. They are instances of dynamically loaded classes based on the descriptor formats, designed to process, validate, or manage specific data. Each handler can implement complex logic and maintain state, depending on the requirements of the associated format. This approach ensures modularity, configurability, and extensibility, far beyond the role of a simple placeholder.

peppelinux:
@LadyCodesItBetter during the dev call we have decided to include wiuthin this PR the interface with the mandatory set of methods to be implemented by the classes used for the VP parse

we also have discussed about the value to handle a proper generalization of each parser configuration, using *args and **kwargs, and therefore enabling the presence of the parameter `config` with each handler configuration

peppelinux (changes requested):
class interance and configuration generalization are required

LadyCodesItBetter:
> class interance and configuration generalization are required

@peppelinux should be done

pyeudiw/tests/openid4vp/test_presentation_submission.py
```diff
@@ -0,0 +1,85 @@
...
+    "MAX_SUBMISSION_SIZE": 10 * 1024  # 10 KB
+}
+
+valid_submission = {
+    "id": "submission_id",
+    "definition_id": "definition_id",
+    "descriptor_map": [
+        {"id": "descriptor_1", "format": "ldp_vp", "path": "$"},
+        {"id": "descriptor_2", "format": "jwt_vp_json", "path": "$"}
+    ]
```
peppelinux, line 23: there paths doesn't really work

````
{
  "id": "Presentation example 2",
  "definition_id": "Example with multiple VPs",
  "descriptor_map": [
    {
      "id": "ID Card with constraints",
      "format": "ldp_vp",
      "path": "$[0]",
      "path_nested": {
        "format": "ldp_vc",
        "path": "$.verifiableCredential[0]"
      }
    },
    {
      "id": "Example credential disclosing only address",
      "format": "dc+sd-jwt",
      "path": "$[1]"
    }
  ]
}
````

File: pyeudiw/openid4vp/presentation_submission/config.yml [hunk 11]
```diff
@@ -0,0 +1,15 @@
+formats:
+  - name: "dc+sd-jwt"
+    module: "pyeudiw.openid4vp.presentation_submission"
+    class: "VcSdJwt"
```
peppelinux, line 4: https://github.com/italia/eudi-wallet-it-python/blob/dev/pyeudiw/sd_jwt/verifier.py

```suggestion
    module: "pyeudiw.sd_jwt.verifier"
    class: "SDJWTVerifier"
```

PR #321, opened 2025-01-09
Feature/sd jwt issuer verify iat exp

### LINKED ISSUE #316: SD-JWT for VC should use JWSHelper to verify the issuer JWT
Currently, the SD-JWT for VC module tries to recover the issuer public key of the SD-JWT.
(1) First, the module search for a valid key indentifier from the jwt header (kid, x5c, trust_chain, etc).
https://github.com/italia/eudi-wallet-it-python/blob/766ba3fb469947bd1c7d40634e9819d90f080ac1/pyeudiw/openid4vp/vp_sd_jwt_vc.py#L33-L34
(2) then it tries to match the identifier with the available public keys whitelisted from the trust evaluation mehcanism, here
https://github.com/italia/eudi-wallet-it-python/blob/766ba3fb469947bd1c7d40634e9819d90f080ac1/pyeudiw/tools/jwk_handling.py#L6-L30
(3) Finally, after a candidate key is identified, it is then used for the issuer jwt verification, here
https://github.com/italia/eudi-wallet-it-python/blob/766ba3fb469947bd1c7d40634e9819d90f080ac1/pyeudiw/openid4vp/vp_sd_jwt_vc.py#L43-L44

However, after PR #315, the flow above has been replicated and stabilized in the JWSHelper class, which now will extract (or match) the valid verifying key form the token header with one of the whitelisted set provided in the class constructor.
https://github.com/italia/eudi-wallet-it-python/blob/766ba3fb469947bd1c7d40634e9819d90f080ac1/pyeudiw/jwt/__init__.py#L376-L395

As such, the SD-JWT for VC module should be refactored to use the new JWSHelper in order to remove the duplicated code and responsability concerning the identification of the issuer public key and its usage for token verification.

For history and completeness: this issue here described is linked with https://github.com/italia/eudi-wallet-it-python/pull/305#issuecomment-2478525959 and a refactoring to use JWSHelper would also solve the the problems declared in that comments.

LadyCodesItBetter:
Resolve #316

### DIFF
--- a/pyeudiw/jwt/jws_helper.py
+++ b/pyeudiw/jwt/jws_helper.py
@@ -230,7 +230,7 @@ class JWSHelper(JWHelperInterface):
         
         # Verify the JWS compact signature
         verifier = JWS(alg=header["alg"])
-        msg = verifier.verify_compact(jwt, [key_from_jwk_dict(verifying_key)])
+        msg: dict = verifier.verify_compact(jwt, [key_from_jwk_dict(verifying_key)])
 
         # Validate JWT claims
         try:
--- a/pyeudiw/jwt/jws_helper.py
+++ b/pyeudiw/jwt/jws_helper.py
@@ -27,21 +27,6 @@ DEFAULT_SIG_KTY_MAP = {
     "EC": "ES256"
 }
 
-DEFAULT_SIG_ALG_MAP = {
-    "RSA": "RS256",
-    "EC": "ES256"
-}
-
-DEFAULT_ENC_ALG_MAP = {
-    "RSA": "RSA-OAEP",
-    "EC": "ECDH-ES+A256KW"
-}
-
-DEFAULT_ENC_ENC_MAP = {
-    "RSA": "A256CBC-HS512",
-    "EC": "A256GCM"
-}
-
 class JWSHelper(JWHelperInterface):
     """
     Helper class for working with JWS, extended to support SD-JWT.
@@ -98,10 +83,6 @@ class JWSHelper(JWHelperInterface):
             protected = {}
         if unprotected is None:
             unprotected = {}
-            
-        # Add SD-JWT claims if the payload matches the criteria
-        if isinstance(plain_dict, dict) and self._is_sd_jwt_payload(plain_dict):
-            plain_dict = self._add_sd_jwt_claims(plain_dict)
 
         # Select the signing key
         signing_key = self._select_signing_key((protected, unprotected), signing_kid)  # TODO: check that singing key is either private or symmetric
@@ -110,7 +91,7 @@ class JWSHelper(JWHelperInterface):
         header_kid = protected.get("kid")
         signer_kid = signing_key.get("kid")
         if header_kid and signer_kid and (header_kid != signer_kid):
-            raise JWSSigningError(f"Token header contains kid {header_kid}, which does not match the signing key kid {signer_kid}.")
+            raise JWSSigningError(f"token header contains a kid {header_kid} that does not match the signing key kid {signer_kid}")
 
         payload = serialize_payload(plain_dict)
         
@@ -134,14 +115,19 @@ class JWSHelper(JWHelperInterface):
             signing_key.pop("kid", None)
 
         signer = JWS(payload, alg=signing_alg)
-        
         if serialization_format == "compact":
             try:
-                signed = signer.sign_compact([key_from_jwk_dict(signing_key)], protected=protected, **kwargs)
+                signed = signer.sign_compact(
+                    [key_from_jwk_dict(signing_key)], protected=protected, **kwargs
+                )
                 return signed
             except Exception as e:
                 raise JWSSigningError("Signing error: error in step", e)
-        return signer.sign_json(keys=[key_from_jwk_dict(signing_key)], headers=[(protected, unprotected)], flatten=True)
+        return signer.sign_json(
+            keys=[key_from_jwk_dict(signing_key)],
+            headers=[(protected, unprotected)],
+            flatten=True,
+        )
 
     def _select_signing_key(self, headers: tuple[dict, dict], signing_kid: str = "") -> dict:
         if len(self.jwks) == 0:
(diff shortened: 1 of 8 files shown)


pyeudiw/jwt/jws_helper.py
```diff
@@ -134,14 +131,19 @@ def sign(
...
-                signed = signer.sign_compact([key_from_jwk_dict(signing_key)], protected=protected, **kwargs)
+                signed = signer.sign_compact(
+                    [key_from_jwk_dict(signing_key)], protected=protected, **kwargs
+                )
                 return signed
             except Exception as e:
                 raise JWSSigningError("Signing error: error in step", e)
-        return signer.sign_json(keys=[key_from_jwk_dict(signing_key)], headers=[(protected, unprotected)], flatten=True)
+        return signer.sign_json(
+            keys=[key_from_jwk_dict(signing_key)],
```
peppelinux, line 143: don't repeat this twice `key_from_jwk_dict(signing_key)`

please assign the result to a variable and reuse it

pyeudiw/jwt/jws_helper.py
```diff
@@ -223,32 +225,18 @@ def verify(self, jwt: str) -> (str | Any | bytes):
                         "unexpected verification state: found a valid verifying key,"
                         f"but its kid {obtained_kid} does not match token header kid {expected_kid}")
                 )
-
-        verifier = JWS(alg=header["alg"])
-        msg = verifier.verify_compact(jwt, [key_from_jwk_dict(verifying_key)])
         
         # Verify the JWS compact signature
         verifier = JWS(alg=header["alg"])
         msg = verifier.verify_compact(jwt, [key_from_jwk_dict(verifying_key)])
```
peppelinux, line 231:
```suggestion
        msg :dict = verifier.verify_compact(jwt, [key_from_jwk_dict(verifying_key)])
```

pyeudiw/jwt/jws_helper.py
```diff
@@ -270,40 +258,48 @@ def _select_verifying_key(self, header: dict) -> dict | None:
...
             return False
-        return payload.get("typ") == "sd-jwt"
-    
-    def _add_sd_jwt_claims(self, payload: dict) -> dict:
-        """
-        Adds SD-JWT specific claims to the payload.
 
-        :param payload: The original payload.
-        :returns: The payload with added SD-JWT claims.
+    def validate_jwt_claims(self, payload: dict) -> None:
```
peppelinux, line 285: this would be renamed to something specific for the timestamps used in iat, exp and nbt

`validate_jwt_timestamps_claims`

pyeudiw/sd_jwt/verifier.py
```diff
@@ -112,15 +114,41 @@ def _verify_sd_jwt(
...
+                raise JWSVerificationError(f"Invalid JWT claims: {e}")
+            
         else:
             raise ValueError(
                 f"Unsupported serialization format: {self._serialization_format}"
             )
         
         self._holder_public_key_payload = self._sd_jwt_payload.get("cnf", None)
 
+    def validate_claims(self, payload: dict) -> None:
```
peppelinux, line 130: the same code is repeated here: https://github.com/italia/eudi-wallet-it-python/pull/321/files#diff-0317cdc2af0d2f80fb9f878259e22ed628fea69148372f976a757b8bee5b0c6aR285

pyeudiw/jwt/jws_helper.py
```diff
@@ -120,7 +117,7 @@ def sign(
         
          # Add "typ" header if not present
         if "typ" not in protected:
-            protected["typ"] = "sd-jwt" if self._is_sd_jwt_payload(plain_dict) else "JWT"
+            protected["typ"] = "sd-jwt" if self.is_sd_jwt(plain_dict) else "JWT"
```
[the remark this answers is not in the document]
Zicchio: `plain_dict` is actually the token payload (as a dictionary) to be signed and passed in the function argument; see
https://github.com/italia/eudi-wallet-it-python/blob/4267f0310bc5381c458f49d9c72e7bdd32e397a8/pyeudiw/jwt/jws_helper.py#L83

it's not a very clear name but it's what the repo always used (I'm not sure why): see this reference on the helper from >1 year ago
https://github.com/italia/eudi-wallet-it-python/blob/c6a81ca86fcd222f40903cfb8fe3e87aec5d06ae/pyeudiw/jwt/__init__.py#L106-L111

pyeudiw/sd_jwt/verifier.py
Zicchio: Moreover, note that the method does not need to be class member: `self` is never used and this could be a pure function instead (ivoking the pure function twice also trivially satisfy the review note above).
IMHO I would name the method `validate_lifetime_claims(claims: dict)`, but `valida_timestamp_claims` is also 100% ok.

pyeudiw/sd_jwt/verifier.py
```diff
@@ -112,15 +114,41 @@ def _verify_sd_jwt(
...
+
+        :param payload: The decoded JWT payload.
+        :type payload: dict
+        :raises ValueError: If any of the claims are invalid.
+        """
+        current_time = iat_now()
+
+        if 'iat' in payload:
+            if payload['iat'] > current_time:
+                raise ValueError("Future issue time, token is invalid.")
```
Zicchio, line 142: I'm not sure it matters but this is not a ValueError. This is a *validation* error. In Python a ValueError is usually used when a passed argument has a valid type, but the method in undefined on that specific value (for example a method could be defined on `int` but might only be able to process positive integers).
https://docs.python.org/3/library/exceptions.html#ValueError

A custom exception, such as `LifetimeException` or something like that, might be more suitable.

pyeudiw/jwt/jws_helper.py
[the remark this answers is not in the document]
peppelinux: since it is the sign method, payload alone would be a better name

pyeudiw/sd_jwt/verifier.py
peppelinux: LifetimeException that inherits ValidationError

pyeudiw/jwt/jws_helper.py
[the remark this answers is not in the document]
LadyCodesItBetter: I agree that the name plain_dict can be confusing, as also pointed out by @peppelinux and confirmed by @Zicchio .

I would suggest considering the possibility of updating the parameter name in the future to improve code readability, for instance, using a term like payload. This would make the meaning more intuitive for anyone reading the code, especially those unfamiliar with the repository's historical implementation.

pyeudiw/jwt/jws_helper.py
LadyCodesItBetter: Ok

peppelinux (approved):

Issue #310, opened 2024-11-20
[Interoperability] Support for unencrypted direct_post authz response mode 


Zicchio:
Currently in this project, in the opanid4vp flow we provide support exclusively for `direct_post.jwt` encrypted authorization response (accordingly with italian specs).

It appears that, as of today, the only authorization response mode supported in Potential interoperability D2.2 is the (unecrnytped) `direct_post` mode.
It is not clear if or when this requirement will be relaxed to also include encrypted `direct_post.jwt` mode, hence **if** we want to comply with interoperability goals we might need to support `direct_post` mode and provide configurations in order to enable it.

I think that no immediate action is required, but I still want to highlighting this fact so that one day we are ready to develope this feature if its necessity actually emerges.

peppelinux:
@Zicchio the solution for a proper interoperability is to always encryopt the response if the RP provide the public keys intended fro the encryption usage

the normative language in the italian specs should mandate the RP to have encryption JWKs in their metadata, while wallet units should try to encrypt if the rp allow them to do so

for our implementation we therefore newed to do encryption if the RP provide the public enc material suitable for this

peppelinux:
for instance, in our implementation the RP configures its private keys (and intended usage) here:
https://github.com/italia/eudi-wallet-it-python/blob/dev/example/satosa/pyeudiw_backend.yaml#L136

Zicchio:
So, if I understand correctly, we require that the authorization response MUST be encrypted **IF AND ONLY IF** the relying party has a configured (public) encryption key.

Looks good to me.

Zicchio:
After a discussion, the approach on how to solve this problem has changed. Instead, we support both. To do so, make inference on the response body (giving priority to jwt) in this block of code

https://github.com/italia/eudi-wallet-it-python/blob/68246ab8358d10782109a44179699c5eb87e3103/pyeudiw/satosa/default/response_handler.py#L149-L152

Zicchio:
> After a discussion, the approach on how to solve this problem has changed. Instead, we support both. To do so, make inference on the response body (giving priority to jwt)

PR #325 add the option to configure the authorization request to set `direct_post` instead of `direct_post.jwt (this is in accordance with potential-lsp, see #317).
[...]

PR #325, opened 2025-01-14
Expand Request Object configurability 

### LINKED ISSUE #290: [Satosa] Rename metadata `presentation_definition` to `presentation_definitions_supported`
The satosa metadata configuration include a field called `presentation_definition` but in the current version of the specs it is called `presentation_definitions_supported`.
For further references, see
https://github.com/italia/eudi-wallet-it-docs/blob/fdd9cf44e69bd6eeb4cfb11684dd6f5e0e7c6a57/docs/en/relying-party-entity-configuration.rst?plain=1#L45-L46

### LINKED ISSUE #317: [Interoperability][JAR] Authorization Request Object requirements for interoperability
The interoperability requirements for the Authorization Request (which, in the case of the cross device flow, is the content of the QR code) has some differences with the current implementation. If we want to comply with potential, then we should find a way to satisfy those requirements.

Currently, the differences in the request object compared to the current implementation are:
 - [ ] the request object jwt should include an `aud` claim with value equal to `"https://self-issued.me/v2"`
 - [ ] `client_id_scheme` is omitted as no trust framework is used in potential
 - [ ] `presentation_definition` is inserted directly in the request object as no trust framework is used in potential
 - [ ] `response_mode` should be `direct_post` (this is related to https://github.com/italia/eudi-wallet-it-python/issues/310) as no trust framework is used in protential

I leave myself the option to edit the message is order to include further requirements relating the request object if or when they are observed.

Zicchio:
This PR contians the follwoing features:
1. Expand the request object configurability with `aud` (new optional claim) and `response_mode` (now optional configurable claim, defaults to `direct_post.jwt`).
2. Moved `presentation_definition` from federation metadata (and its configuration) to the request object.
3. Presentation definition example is now an example from potential interoperability document (opain4vp for sd-jwt, v1.1)

This closes #317 with the caveat that `client_id_scheme` is not removed (with the rationale that jwt receiver should omit claims that it cannot understand).
This also closes #290 as a side effect of point 2.

Note that `direct_post` is not fully support. It is JUST configurable. Full support will come when dealing with #310


example/satosa/pyeudiw_backend.yaml
```diff
@@ -62,10 +62,41 @@ config:
...
   authorization:
     url_scheme: haip
-    scopes: 
+    scopes: # deprecated
     - pid-sd-jwt:unique_id+given_name+family_name
     default_acr_value: https://www.spid.gov.it/SpidL2
     expiration_time: 5 # minutes
+    aud: https://self-issued.me/v2
+    response_mode: direct_post
+    presentation_definition:
```
peppelinux, line 71: for now it is good

at the same time, when dcql is defined only... do we suppose to present them all?
I suppose that only on them at time can be brougth in the request

pyeudiw/openid4vp/authorization_request.py
```diff
@@ -13,3 +17,40 @@ def build_authorization_request_url(scheme: str, params: dict) -> str:
...
+        "response_mode": authorization_config.get("response_mode", ResponseMode.direct_post_jwt),
+        "response_type": "vp_token",
+        "response_uri": response_uri,
+        "nonce": str(uuid.uuid4()),
+        "state": state,
+        "iss": client_id,
+        "iat": iat_now(),
+        "exp": exp_from_now(minutes=authorization_config["expiration_time"])
+    }
+
```
peppelinux, line 53:
```suggestion
    claims = {
        "client_id_scheme": "entity_id",  # that's federation.
        "client_id": client_id,
        "presentation_definition": authorization_config["presentation_definition"],
        "response_mode": authorization_config.get("response_mode", ResponseMode.direct_post_jwt),
        "response_type": "vp_token",
        "response_uri": response_uri,
        "nonce": str(uuid.uuid4()),
        "state": state,
        "iss": client_id,
        "iat": iat_now(),
        "exp": exp_from_now(minutes=authorization_config["expiration_time"])
    }
    if authorization_config.get("scopes"):
        claims["scopes"] = authorization_config["scopes"]
```

peppelinux: entity_id does not exist anymore, it is now http

peppelinux (approved):

example/satosa/pyeudiw_backend.yaml
Zicchio: presentation definition and dcql are indeed exclusive: in an authorization requeste, only one among `presentation_definition` and `dcql_query` can be defined
https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#name-new-parameters

DCQL also changes the shape of the response, which (1) does not include the presentation submission and (2) changes the vp_token structure.
https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#name-response-parameters

DCQL will indeed include a non-trivial amount of changes in the code...

pyeudiw/openid4vp/authorization_request.py
```diff
@@ -13,3 +17,40 @@ def build_authorization_request_url(scheme: str, params: dict) -> str:
...
+    """
+    claims = {
+        "scope": ' '.join(authorization_config["scopes"]),
+        "client_id_scheme": "entity_id",  # that's federation.
+        "client_id": client_id,
+        "presentation_definition": authorization_config["presentation_definition"],
+        "response_mode": authorization_config.get("response_mode", ResponseMode.direct_post_jwt),
+        "response_type": "vp_token",
+        "response_uri": response_uri,
+        "nonce": str(uuid.uuid4()),
```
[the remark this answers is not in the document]
Zicchio: nonce in indirectly saved in the DB as a side effect of the fact the whole request object payload is saved in the DB (we use it later on when doing holder-key-binding validation, so we need it).

https://github.com/italia/eudi-wallet-it-python/blob/1ed8b65cd9832529c0781b46f1a9855a35269702/pyeudiw/satosa/default/request_handler.py#L49

(I also has this "wait a second" moment when refactoring this function and observing that the nonce was generated and not *directly* saved anywhere: it took me a while to remember what happened in code when it tried to recover the nonce)

example/satosa/pyeudiw_backend.yaml
Zicchio: Added nonce as optional arugment in commit 2a85c407caa252af94dd7192b9e25d7d0bc3383c

pyeudiw/openid4vp/authorization_request.py
Zicchio: updated as suggest with commit 2a85c407caa252af94dd7192b9e25d7d0bc3383c, except that the claim `scope` in the request object is singular and not plural (as by OAuth RFC).
The idea was that we would configure a scopeS list and than the code would flatten it as space-sperataed string.

That said, why were we configuring `scopes` if we need `scope`? Maybe open an issue for this in the future

README.SATOSA.md
```diff
@@ -75,11 +75,28 @@ To install the OpenID4VP SATOSA backend you just need to:
...
+| ------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------- | ---------------------------------------------- |
+| config.authorization.presentation_definition.id                           | The unique identifier of the presentation definition                                                  | d76c51b7-ea90-49bb-8368-6b3d194fc131           |
+| config.authorization.presentation_definition.input_descriptors            | The list of input descriptors that specify the verifiable credentials that the client requests        | See below                                      |
+| config.authorization.presentation_definition.id                           | The unique identifier of the input descriptor                                                         | IdentityCredential                             |
+| config.authorization.presentation_definition.format                       | The object that defines the verifiable credential format that the client requests                     | vc+sd-jwt: {}                                  |
```
peppelinux, line 94:
```suggestion
| config.authorization.presentation_definition.format                       | The object that defines the verifiable credential format that the client requests                     | dc+sd-jwt: {}                                  |
```

example/satosa/pyeudiw_backend.yaml
```diff
@@ -62,10 +62,41 @@ config:
     
   authorization:
     url_scheme: haip
-    scopes: 
+    scopes: # deprecated
```
peppelinux, line 65:
```suggestion
    scopes:
```

example/satosa/pyeudiw_backend.yaml
peppelinux: dcql will be developed in a separate component

peppelinux (approved):

pyeudiw/openid4vp/authorization_request.py
peppelinux: typo, I just have fied it, it is called "nonce"

peppelinux (approved):

PR #328, opened 2025-01-15
feat: publish trust metadata endpoint

### LINKED ISSUE #296: [Satosa][Interoperability]JAR Metadata endpoint: Exposure of RP signing key of request object
In the "POTENTIAL: API Piloting Definition Scope OID4VP V1.0" _draft_ document (which unfortunately I cannot link), regarding the request object JWT (i.e. JAR protocol), it is stated that

>_Within (11) is explained how the client that received JWT (in our case wallet that received JAR) can fetch public key of the issuer (in our case RP). It uses HTTP GET method.
RP will publish its metadata containing jwk at the location formed by inserting the well-known string (for Potential we will use /.well-known/jar-issuer) between the host component and the path component (if any) of the iss claim value in the JAR._

If I am understanding this correctly, this implied that some module (direct trust?) should expose a `/.well-known/jar-issuer` endpoint that exposes the public key whose private component was used to sign the request object JWT.

Asking @peppelinux for:
1. confirmation if what I understood is correct
2. if we should add this requirement somewhere in some milestone (maybe 9.2?)

### LINKED ISSUE #312: metadata endpoints according to the configured trust evalaution mechanisms
register endpoints (https://github.com/italia/eudi-wallet-it-python/blob/dev/pyeudiw/satosa/default/openid4vp_backend.py#L101) should use the endpoints (and related resources) configured for each trust evaluation mechanisms configured within the project

Zicchio:
This PR includes the following:
1. extends the TrustHandler interface to allow the possibility to publish trust-framework defined endpoints
2. introduce a new trust evaluation mechanism, called DirectTrustJar, that extends DirectTrustSdJwtVc, to let RP publish its own signing keys

Notably, it does NOT update the former federation publication mechanism to use this new endpoint. This is because this task is not-trivial and probably requires a dedicated pull request.
With the exception above, this closes #312 and closes #296. When merged, a new issue should be opened to update federation.

Due to restrictions introduced by which endpoint can be exposed by a satosa backend module, integration tests require this modification to work https://github.com/italia/iam-proxy-italia/pull/177 . These two pull request are intended to be accepted side by side.

### DIFF
--- a/example/satosa/integration_test/commons.py
+++ b/example/satosa/integration_test/commons.py
@@ -1,5 +1,7 @@
 import base64
 from copy import deepcopy
+import json
+from urllib.parse import urlparse
 from pyeudiw.tools.utils import exp_from_now, iat_now
 from bs4 import BeautifulSoup
 import datetime
@@ -9,7 +11,7 @@ from typing import Any, Literal
 from io import StringIO
 from pyeudiw.jwk import JWK
 from pyeudiw.jwt.jwe_helper import JWEHelper
-from pyeudiw.jwt.jws_helper import DEFAULT_SIG_KTY_MAP
+from pyeudiw.jwt.jws_helper import DEFAULT_SIG_KTY_MAP, JWSHelper
 from pyeudiw.jwt.utils import decode_jwt_payload
 from pyeudiw.sd_jwt.issuer import SDJWTIssuer
 from pyeudiw.sd_jwt.utils.yaml_specification import _yaml_load_specification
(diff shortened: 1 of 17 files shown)


peppelinux (approved):

Zicchio:
@peppelinux with c0ced4a the commonalities between fetching / exposing a JWK with direct trust JAR and direct trust SD-JWT for VC have been grouped to a common (private) class called `_DirectTrustJwkHandler` (the name is whatever: I am open for suggestions).
In pratical terms, TrustJar and TrustSdJwt currently do nothing but provide sensible defaults for endpoint paths of `_DirectTrustJwkHandler`.

Currently, Both TrustJar and TrustSdJwt INHERIT from `_DirectTrustJwkHandler` but if desired we can migrate to a composition approach in the future (tranisition should be smooth): this would be especially desirable if jar and sd-jwt eventually require management of metadata other than key exposition, but currently this is not case so inheritance should be fine.

PR #332, opened 2025-01-27
Customizability of client id in jar request object

### LINKED ISSUE #331: Customizability of client id in request object
As per Potential D2.2 version 1.1, the client id in the JAR request object should be a string (no colon characters are allowed) with the following pattern `<country>-<RPname>`

elisanp:
This closes #331


pyeudiw/openid4vp/authorization_request.py
```diff
@@ -19,12 +19,12 @@ def build_authorization_request_url(scheme: str, params: dict) -> str:
     return f"{scheme}{_sep}{query_params}"
 
 
-def build_authorization_request_claims(client_id: str, state: str, response_uri: str, authorization_config: dict, nonce: str = "") -> dict:
+def build_authorization_request_claims(token_iss: str, state: str, response_uri: str, authorization_config: dict, nonce: str = "") -> dict:
```
peppelinux, line 22: client_id remains as client_id

it is implict that the signed request object is considered also a token and that the client is its issuer

Zicchio:
Please document the new configuration here
https://github.com/italia/eudi-wallet-it-python/blob/1db0bc592150108dc86a2049d5bc16a8440fb91e/README.SATOSA.md?plain=1#L76
stating that is is optional and only influence the client in the JAR and not the whole project.

pyeudiw/openid4vp/authorization_request.py
Zicchio: In the potential request object, the iss and client ID are different. 
Iss is an HTTPS url value.
Client ID is a string with pattern <country>-<RPname>.

I am clueless on the the why(s) of that choice.

The chhaneg in the method signature should reflect that change.

Zicchio (approved):

README.SATOSA.md
```diff
@@ -77,10 +77,11 @@ To install the OpenID4VP SATOSA backend you just need to:
 
 | Parameter                                    | Description                                                                                         | Example value                                       |
 | -------------------------------------------- | --------------------------------------------------------------------------------------------------- | --------------------------------------------------- |
+| config.authorization.client_id               | Optional client_id to be set only in the request object; if not set it is equal to the issuer       | IT-PYEUDIRP                                         |
```
peppelinux, line 80:
```suggestion
| config.authorization.client_id               | client_id to be set only in the request object; if not set default to entity id value      | IT-PYEUDIRP (as requested by Potential LSP)   |
```

peppelinux (changes requested):
per my comments above

example/satosa/pyeudiw_backend.yaml
```diff
@@ -61,6 +61,7 @@ config:
         - ES512
     
   authorization:
+    client_id: custom-client-id
```
peppelinux, line 64:
```suggestion
    client_id: # this field if not set will be autopopulated using internal variables base_url and name using the following format: "<base_url>/<name>"

**Review by peppelinux — APPROVED** [review]

File: pyeudiw/openid4vp/authorization_request.py [hunk 4]
```diff
@@ -1,69 +1,65 @@
...
+
+    claims = {
+        "client_id_scheme": "http",  # that's federation.
+        "client_id": client_id,
+        "response_mode": authorization_config.get("response_mode", ResponseMode.direct_post_jwt),
+        "response_type": "vp_token",
+        "response_uri": response_uri,
+        "nonce": nonce,
+        "state": state,
+        "iss": authorization_config.get("auth_iss_id", client_id),
```
Zicchio, line 53: Minor note: a problem with this is that it is surprisingly complicated, when using custom client id, to have `iss=f"{base_url}/{name}"` which we had before. To do so, when configuring `pyeudiw.backend.yaml` you have to known what are `base_url` and `name`. The value for `name` is ok (it is already in the config file) but in generale the backend module configuration should not need to know what `base_url` is as it is a satosa configuration and is ultimately decided by the satosa core module.

This can be solved by using some environment variable (is it this one? https://github.com/italia/iam-proxy-italia/blob/e94ef7bee7d92f06159f47367af780251bf1d0e7/example/proxy_conf.yaml#L3C12-L3C23) but IMO it is a weird workaround.

peppelinux (approved):

example/satosa/pyeudiw_backend.yaml
```diff
@@ -61,6 +61,8 @@ config:
         - ES512
     
   authorization:
+    client_id: # this field if not set will be autopopulated using internal variables base_url and name using the following format: "<base_url>/<name>" 
+    auth_iss_id: # this field if not set will be autopopulated using internal variables base_url and name using the following format: "<base_url>/<name>"
```
peppelinux, line 65:
```suggestion
    auth_iss_id: # this field if not set will be set to client_id in the authz request 
```

Issue #349, opened 2025-01-30
Define a ValidationError Exception without depending from pydantic


peppelinux:
````
grep -nHIirF -- from pydantic import ValidationError (nella directory: /home/wert/DEV/DTD/Wallet/eudi-wallet-it-python/pyeudiw)
./satosa/default/response_handler.py:8:from pydantic import ValidationError
./satosa/default/openid4vp_backend.py:5:from pydantic import ValidationError
./satosa/utils/dpop.py:3:from pydantic import ValidationError
./openid4vp/direct_post_response.py:18:from pydantic import ValidationError
./federation/statements.py:16:from pydantic import ValidationError
./tests/openid4vp/schemas/test_wallet_instance_attestation.py:2:from pydantic import ValidationError
./tests/openid4vp/schemas/test_vp_token.py:2:from pydantic import ValidationError
./tests/openid4vp/schemas/test_schema.py:2:from pydantic import ValidationError
./tests/openid4vp/schemas/test_wallet_instance_attestation_request.py:2:from pydantic import ValidationError
./tests/federation/schemas/test_entity_configuration.py:2:from pydantic import ValidationError
./jwt/helper.py:3:from pydantic import ValidationError
Search completed with 11 matches.
````

PR #350, opened 2025-01-30
pydantic.ValidationError exception refinements

peppelinux:
This PR resolves https://github.com/italia/eudi-wallet-it-python/issues/349

- when the validation error is about pydantic, the Exception must be handled using pydantic.ValidationError
- sparse exceptions classes are now collected only within the specialized expcetions definition files
- huge code linting (sorry but sooner or later ...)


pyeudiw/jwt/exceptions.py
```diff
@@ -20,3 +23,7 @@ class JWEEncryptionError(Exception):
 
 class JWTDecodeError(Exception):
     pass
+
+
+class LifetimeException(ValidationError):
```
Zicchio, line 28: I think that IMO this should inherit from `pyeudiw.exceptions.ValidationError` rather than `pydantic.ValidationError`.

Pydantic ValidationError is special, in meant to include a *list* of validation errors (which we don't create), is meant to include the reference model that was not satisfied (which we do not include) and for a long time was not meant to be inherited from. IMO it is likely that inheriting from `pydantic.ValidationError` might result in errors downstream, especially if we use `pydantic.ValidationError` methods (which would fail because we do not satisfy the ValidationError constructor when using LifetimeException).

See
https://github.com/pydantic/pydantic/blob/5e368925f607d47c4d98a709b170ea13615b05f2/pydantic/v1/error_wrappers.py#L50

thomaschiozzi-tndigit (approved):

PR #352, opened 2025-01-31
Huge cleanups and optimizations 

### LINKED ISSUE #334: Migrate federation published endpoint to new approach
The project surrently publish a federation endpoint `/.well-known/openid-federation`. To do so, the project uses a custom ad-hoc solution for federation by appending it in the project-defined `satosa.backends.BackendModule` class attributes; see here
https://github.com/italia/eudi-wallet-it-python/blob/1db0bc592150108dc86a2049d5bc16a8440fb91e/pyeudiw/satosa/utils/trust.py#L201 
, which is then published here
https://github.com/italia/eudi-wallet-it-python/blob/1db0bc592150108dc86a2049d5bc16a8440fb91e/pyeudiw/satosa/default/openid4vp_backend.py#L116

Since PR #328 introduced a standard default machanism to let trust handler publish arbitrary trust-associated endpoints, federation should also be armonized to used this approach in order to not have multiple ways to do the same things in the code base (which unfortunately happens often enough).

### LINKED ISSUE #342: Scout and remove dead code
The codebase is accumulating some amount of dead code. For easyness of use, it should be removed.
Some examples below.

This is dead code EDIT: removed in #347 
https://github.com/italia/eudi-wallet-it-python/blob/486aa592d793e7a9922fe83430ce6b6558eefd45/pyeudiw/satosa/default/response_handler.py#L178-L181

This is dead code EDIT: removed in #347 
https://github.com/italia/eudi-wallet-it-python/blob/486aa592d793e7a9922fe83430ce6b6558eefd45/pyeudiw/jwt/parse.py#L61-L73

This is dead code
https://github.com/italia/eudi-wallet-it-python/blob/486aa592d793e7a9922fe83430ce6b6558eefd45/pyeudiw/satosa/default/response_handler.py#L38-L66

peppelinux:
This PR must be used with iam-proxy-italia branch https://github.com/italia/iam-proxy-italia/pull/185:

-  adds a default_client_id and the possibility for each trust evaluation handler to use different client ids (experimental due to the LSP Potential requirements)
- aims to remove all the duplicated code and overlapping classes born during the refactoring from 0.7 to 0.9, with the scope to simplify the implementations
- extends the configurations and consolidates the example configuration to provide a more appropriate guidance to the implementers
- removes the deprecation warning about pyhdantinc `FieldValidationInfo`
- resolves #334
- Resolves #342 
- ...


peppelinux:
these three should be cleaned up too

image

pyeudiw/federation/http_client.py
peppelinux, line 1: @PascalDR please move this http_client in general tools/utils

PR #366, opened 2025-02-19
Fix Trust params handling

### LINKED ISSUE #303: Dynamic JWT header in the presentation request according to the configured trust evaluation methods
We cannot rely solery on openid federation, in particular when the configuration doesn't have federation in

https://github.com/peppelinux/draft-demarco-oauth-status-assertions/pull/96#issuecomment-2468045558

PascalDR:
Resolves #303

### DIFF
--- a/pyeudiw/satosa/default/request_handler.py
+++ b/pyeudiw/satosa/default/request_handler.py
@@ -59,7 +59,7 @@ class RequestHandler(RequestHandlerInterface, BaseLogger):
         # load all the trust handlers request jwt header parameters, if any
 
         _protected_jwt_headers.update(
-            self.trust_evaluator.get_selfissued_jwt_header_trust_parameters(issuer=self.client_id)
+            self.trust_evaluator.get_jwt_header_trust_parameters(issuer=self.client_id)
         )
 
         #  federation_trust_handler_backend_class: TrustHandlerInterface = (
--- a/pyeudiw/satosa/default/request_handler.py
+++ b/pyeudiw/satosa/default/request_handler.py
@@ -57,8 +57,10 @@ class RequestHandler(RequestHandlerInterface, BaseLogger):
         }
 
         # load all the trust handlers request jwt header parameters, if any
-        self.trust_evaluator.get_selfissued_jwt_header_trust_parameters(issuer=self.client_id)
 
+        _protected_jwt_headers.update(
+            self.trust_evaluator.get_selfissued_jwt_header_trust_parameters(issuer=self.client_id)
+        )
 
         #  federation_trust_handler_backend_class: TrustHandlerInterface = (
         #  self.get_trust_backend_by_class_name("FederationHandler")
--- a/pyeudiw/tests/satosa/test_backend.py
+++ b/pyeudiw/tests/satosa/test_backend.py
@@ -5,6 +5,7 @@ import urllib.parse
 from unittest.mock import Mock, patch
 
 import pytest
+import unittest.mock
 from bs4 import BeautifulSoup
 from cryptojwt.jws.jws import JWS
 from satosa.context import Context
@@ -35,7 +36,7 @@ from pyeudiw.tests.settings import (
     WALLET_INSTANCE_ATTESTATION,
 )
 from pyeudiw.trust.handler.interface import TrustHandlerInterface
-from pyeudiw.trust.model.trust_source import TrustSourceData
+from pyeudiw.trust.model.trust_source import TrustSourceData, TrustParameterData
 
 
 class TestOpenID4VPBackend:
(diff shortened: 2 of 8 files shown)


pyeudiw/satosa/default/request_handler.py
```diff
@@ -57,8 +57,10 @@ def request_endpoint(self, context: Context, *args) -> Response:
         }
 
         # load all the trust handlers request jwt header parameters, if any
-        self.trust_evaluator.get_selfissued_jwt_header_trust_parameters(issuer=self.client_id)
 
+        _protected_jwt_headers.update(
+            self.trust_evaluator.get_selfissued_jwt_header_trust_parameters(issuer=self.client_id)
```
peppelinux, line 62: different trust evaluation handler may have different client ids (unfortunately)

we have the default_client_id that sets each trust ahdnler client id automatically if not differently provided

this means that we should be able to customize a client id within each trust handler configurations

PascalDR: An initial support for this feature can be found here:
https://github.com/italia/eudi-wallet-it-python/pull/366/commits/24ae4890fcfea76e1693f024248f5244a49f5dfe

peppelinux (approved):

pyeudiw/tests/satosa/test_backend.py
```diff
@@ -629,6 +630,97 @@ def test_request_endpoint(self, context):
...
+            plain_dict=WALLET_INSTANCE_ATTESTATION,
+            protected={
+                "trust_chain": trust_chain_wallet,
+                "x5c": [],
+            },
+        )
+
+        dpop_wia = wia
+
+        dpop_proof = DPoPIssuer(
```
Zicchio, line 653: I guess there is no harm in including a DPoP header but note that the request endpoint no longer process it as, unless I am mistaken, a DPoP with WIA is not mandated by reference specs anymore

pyeudiw/tests/satosa/test_backend.py
```diff
@@ -629,6 +630,97 @@ def test_request_endpoint(self, context):
...
+        context.http_headers = dict(
+            HTTP_AUTHORIZATION=f"DPoP {dpop_wia}", HTTP_DPOP=dpop_proof
+        )
+
+        context.qs_params = {"id": state}
+
+        # put a trust attestation related itself into the storage
+        # this then is used as trust_chain header parameter in the signed
+        # request object
+        db_engine_inst = DBEngine(CONFIG["storage"])
```
Zicchio, line 668: I am not sure about this step.

What I am understanding is that we are "forcing" the database to have the correct trust attestation in it.
How can the satosa+pyeudiw software up and running reach this state where the trust attestation (about myself) is in the DB? Shouldn't it be naturally achieved when we boot the opeid4vp backend?
Maybe this already happens. If so, it would be grateful to guide me to the corresponding block(s) of code.

pyeudiw/trust/dynamic.py
```diff
@@ -54,7 +54,7 @@ def _retrieve_trust_source(self, issuer: str) -> Optional[TrustSourceData]:
             return None
 
     def _upsert_source_trust_materials(
-        self, issuer: str, trust_source: Optional[TrustSourceData]
+        self, trust_source: Optional[TrustSourceData], issuer: Union[str, None] = None
```
Zicchio, line 57: Having both `Optional[T]` and `Union[T, None]` in the same type hint is unusual when, in Python, `Optional[T]` is simply an alias for `Union[T, None]` (no more, no less: it does not imply that an argument is optional: https://docs.python.org/3/library/typing.html#typing.Optional )

pyeudiw/trust/handler/interface.py
```diff
@@ -101,3 +101,13 @@ def name(self) -> str:
         :rtype: str
         """
         return self.__class__.__name__
+    
+    @property
+    def default_client_id(self) -> str:
```
Zicchio, line 106: I have some question here:
(1) what is a default_client_id?
(2) what is its use caes?
(3) why is it required?

Some guidance or docs would be grateful as it not mentioned in issue #303, which this PR solves (maybe it is already tackled in PR #369 which I haven't read yet; if so please just redirect me there)

pyeudiw/trust/dynamic.py
```diff
@@ -164,7 +167,7 @@ def get_policies(self, issuer: str) -> dict[str, any]:
 
         return trust_source.policies
 
-    def get_selfissued_jwt_header_trust_parameters(self, issuer: str) -> list[dict]:
+    def get_selfissued_jwt_header_trust_parameters(self, issuer: Union[str, None] = None) -> list[dict]:
```
Zicchio, line 170: Probably the function name or the comment are not correct.

The function name says get_**SELFISSUED**_blabla suggesting that the pyeduiw module is the only intended issuer.
On the other hand, the function comment says "of a certain issuer", suggesting that it could just be the part of a trust source that can be found in a jwt token header of _any_ issuer.

What is the intended behaviour?

pyeudiw/trust/handler/interface.py
PascalDR: See: https://github.com/italia/eudi-wallet-it-python/pull/366#discussion_r1963232537

Zicchio: Ok to me

Maybe consider reporting this info in #369 if not already there

Zicchio (approved):

PR #371, opened 2025-02-26
Better trust materials handling 

### LINKED ISSUE #340: Trusted Keys mongo database cache is misbehaving
When recoving trusted keys from a credential issuer, in the original deisgn the keys where cached at the database level as a FALLBACK strategy for situations where they cannot be remotedly fetched (for example, the issuer is temporarely offline ad the `/.well-known/openid-federation` or `/.well-known/jwt-vc-issuer` endpoint).

However, the current implementation does not use the cache as fallback, but as a primary recovery strategy. As such, if the creddential issuer changes signing key, we are unable to find the new remote key until the cache is manually flushed from the mongo database.

This is the bad code: when looking at the trusted material, first we look at the DB (line 85) and then interrogate the trust evaluation mechanism only when this is not possible

https://github.com/italia/eudi-wallet-it-python/blob/486aa592d793e7a9922fe83430ce6b6558eefd45/pyeudiw/trust/dynamic.py#L75-L90

PascalDR:
Partially resolve #340 

This PR implements two possible scenarios:
- The first that is the scenario described from @Zicchio where the trust parameters are always fetched and, if the server is down for some reason, it retrieves the materials from cache. 
- The second is the cache first where the trust materials are retrieved from the cache if already present and not expired or fetched using the trustability protocol.

Both of this scenario can still have security problems for the stored data that could be expired or revoked.
For this reason a force_update option and the revoke method are added to better handle this two cases.

### DIFF
--- a/docs/TRUST.md
+++ b/docs/TRUST.md
@@ -4,6 +4,16 @@ Trust module main responsability is to provide cryptographic material, metadata 
 Users can define their own trust module by realizing and configuring a class that satisfy the interface [pyeudiw.trust.interface.TrustEvaluator](/pyeudiw/trust/interface.py).
 This project includes some default implementation of trust, whose configuration are described below.
 
+## Caching modes
+There are two caching modes that can be used to store the cryptographic material of the parties involved in the protocol.
+- update_first: The cryptographic material is fetched using the handler protocol and then stored in the cache.
+                If the retrieval fails, the cache is used.
+- cache_first: The cryptographic material is fetched using the cache.
+               If the cache is empty, the handler protocol is used to retrieve the cryptographic material and then stored in the cache and returned.
+
+update_first is the default caching mode.
+You can set the caching mode by setting the variable trust_caching_mode in the configuration file.
+
 ## Configuration of default Trust modules
 
 ### Direct Trust for SD-JWT VC
--- a/docs/TRUST.md
+++ b/docs/TRUST.md
@@ -1,8 +1,6 @@
 # Trust 
 
-Trust module main responsability is to provide cryptographic material, metadata and revocation status of parties involved in the [OpnedID4VP](https://openid.net/specs/openid-4-verifiable-presentations-1_0.html) protocol.
-Users can define their own trust module by realizing and configuring a class that satisfy the interface [pyeudiw.trust.interface.TrustEvaluator](/pyeudiw/trust/interface.py).
-This project includes some default implementation of trust, whose configuration are described below.
+# Trust Module
 
 ## Caching modes
 There are two caching modes that can be used to store the cryptographic material of the parties involved in the protocol.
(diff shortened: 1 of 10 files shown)


docs/TRUST.md
```diff
@@ -4,6 +4,16 @@ Trust module main responsability is to provide cryptographic material, metadata
...
 
+## Caching modes
+There are two caching modes that can be used to store the cryptographic material of the parties involved in the protocol.
+- update_first: The cryptographic material is fetched using the handler protocol and then stored in the cache.
+                If the retrieval fails, the cache is used.
+- cache_first: The cryptographic material is fetched using the cache.
+               If the cache is empty, the handler protocol is used to retrieve the cryptographic material and then stored in the cache and returned.
+
+update_first is the default caching mode.
+You can set the caching mode by setting the variable trust_caching_mode in the configuration file.
```
peppelinux, line 15: is trust_caching_mode a configuration parameter configurable within each trust evaluation handler?

PascalDR: No it's global.

peppelinux (approved):

Zicchio:
>For this reason a force_update option and the revoke method are added to better handle this two cases.

I don't think this PR solves the security issue highlighted here https://github.com/italia/eudi-wallet-it-python/issues/340#issuecomment-2621059431

This is because force_update "pushes" the problem to the trust handler, forcing an update
https://github.com/italia/eudi-wallet-it-python/blob/2dadd24d6ed011e2d3ea8d42825af36b7652870e/pyeudiw/trust/dynamic.py#L76-L78

But then the Trust Handler has no way to know which keys in the TrustSource should be flushed. Stale trust parameters can be updated but old keys in the trust source stay there.
One idea might be to flush *all* keys/trust_source from the trust source, then re-evaluate all the trust parameters+keys, but first we should double check if has some unintended side effect. I'm not confident that this solution does not have a bad effect somewhere else.

For what it's worth I think that not solving the problem now is fine. I think that my comment in issue #340 should be de-scoped from this PR and we should eventually open a new issue.

pyeudiw/trust/dynamic.py
```diff
@@ -49,40 +55,87 @@ def _retrieve_trust_source(self, issuer: str) -> Optional[TrustSourceData]:
...
+        entity_id = issuer or "__internal__"
+
+        if not trust_source:
+            trust_source = TrustSourceData.empty(entity_id)
+
+        if entity_id == "__internal__":
+            return self._cache_upsert_source_trust_materials(trust_source, issuer)
+        
+        if self.mode == "update_first" or force_update:
+            return self._update_upsert_source_trust_materials(trust_source, issuer)
```
Zicchio, line 134: I have a question: this method calls extract_and_update_trust_materials, which usually (among the other thing) fetches a key and pushes it to the trust source
https://github.com/italia/eudi-wallet-it-python/blob/2dadd24d6ed011e2d3ea8d42825af36b7652870e/pyeudiw/trust/handler/_direct_trust_jwk.py#L196
in the current implementation of add_key this is just a list append
https://github.com/italia/eudi-wallet-it-python/blob/2dadd24d6ed011e2d3ea8d42825af36b7652870e/pyeudiw/trust/model/trust_source.py#L113

Would this lead to a scenario where the trust source has an ever-growing list of duplicate keys?

peppelinux: Duplocates should not be added

Good point

PascalDR:
> > For this reason a force_update option and the revoke method are added to better handle this two cases.
> 
> I don't think this PR solves the security issue highlighted here [#340 (comment)](https://github.com/italia/eudi-wallet-it-python/issues/340#issuecomment-2621059431)
> 
> This is because force_update "pushes" the problem to the trust handler, forcing an update
> 
> https://github.com/italia/eudi-wallet-it-python/blob/2dadd24d6ed011e2d3ea8d42825af36b7652870e/pyeudiw/trust/dynamic.py#L76-L78
> 
> But then the Trust Handler has no way to know which keys in the TrustSource should be flushed. Stale trust parameters can be updated but old keys in the trust source stay there. One idea might be to flush _all_ keys/trust_parameters from the trust source, then re-evaluate all the trust parameters+keys, but first we should double check if has some unintended side effect. I'm not confident that this solution does not have a bad effect somewhere else.
> 
> For what it's worth I think that not solving the problem now is fine. I think that my comment in issue #340 should be de-scoped from this PR and we should eventually open a new issue.

I said partially for this reason.
We need to address the remaining point in separate issue PR.

pyeudiw/trust/model/trust_source.py
```diff
@@ -112,21 +112,28 @@ def __init__(
...
     def add_key(self, key: dict) -> None:
         """
-        Add a key to the trust source.
+        Add a key to the trust source if it is not already present.
 
         :param key: The key to add
         :type key: dict
         """
-        self.keys.append(key)
+        filtered_keys = [k for k in self.keys if k["kid"] == key["kid"]]
```
Zicchio, line 120: this filtering logic might not be sound as unfortunately kid is optional and provides no uniqueness guarantees according to [RFC7517](https://datatracker.ietf.org/doc/html/rfc7517#section-4.5)

pyeudiw/trust/model/trust_source.py
```diff
@@ -112,21 +112,29 @@ def __init__(
...
         """
-        Add a key to the trust source.
+        Add a key to the trust source if it is not already present.
 
         :param key: The key to add
         :type key: dict
         """
-        self.keys.append(key)
+        key_thumbprint = key_from_jwk_dict(key).thumbprint("SHA-256")
+        filtered_keys = [k for k in self.keys if key_from_jwk_dict(key).thumbprint("SHA-256") == key_thumbprint]
```
Zicchio, line 121: typo/coding error
`key_from_jwk_dict(key).thumbprint("SHA-256") == key_thumbprint` trivially always evaluates to `True`
i think you meant
`key_from_jwk_dict(k).thumbprint("SHA-256") == key_thumbprint`

Zicchio (approved):

PR #375, opened 2025-03-03
feat: parsing ec key from x5c claim

### LINKED ISSUE #343: Support x5c chains terminating with an EC key
Current strategy for "self contained" key extraction from jwt header does not support x5c chain that represent an EC key. Only RSA is supproted.

https://github.com/italia/eudi-wallet-it-python/blob/486aa592d793e7a9922fe83430ce6b6558eefd45/pyeudiw/jwk/parse.py#L21-L29

This would be the fixing code (that should be tested)
```
def parse_key_from_x5c(x5c: list[str]) -> JWK:
    """Parse a key from an x509 chain. This function currently
    support only the parsing of public RSA and EC key from such a chain.
    The first element of the chain will contain the verifying key.
    See RFC7517 https://datatracker.ietf.org/doc/html/rfc7517#section-4.7
    """
    try:
        # maybe RSA?
        public_key = import_rsa_key(x5c[0])
        key_dict = RSAKey(pub_key=public_key).to_dict()
        return JWK(key_dict)
    except Exception:
        # maybe EC?
        public_key = import_ec_key(x5c[0])
        key_dict = ECKey(pub_key=public_key).to_dict()
        return JWK(key_dict)
    except Exception:
        # neither RSA nor EC
        raise InvalidJwk(f"unable to parse key from x5c: {x5c}")
```

Zicchio:
This pull requests closes #343 

Currently missing an unit test akin to https://github.com/italia/eudi-wallet-it-python/blob/dev/pyeudiw/tests/jwk/test_parse.py

### DIFF
--- a/pyeudiw/tests/satosa/test_backend.py
+++ b/pyeudiw/tests/satosa/test_backend.py
@@ -594,7 +594,7 @@ class TestOpenID4VPBackend:
                     {
                         "id": "pid-sd-jwt:unique_id+given_name+family_name",
                         "path": "$.vp_token.verified_claims.claims._sd[0]",
-                        "format": "vc+sd-jwt"
+                        "format": "dc+sd-jwt"
                     }
                 ]
             }
--- a/pyeudiw/jwk/parse.py
+++ b/pyeudiw/jwk/parse.py
@@ -3,25 +3,26 @@ from cryptojwt.jwk.rsa import RSAKey, import_rsa_key
 
 from pyeudiw.jwk import JWK
 from pyeudiw.jwk.exceptions import InvalidJwk
+from typing import Optional
 
-
-def parse_key_from_x5c(x5c: list[str]) -> JWK:
-    """Parse a key from an x509 chain. This function currently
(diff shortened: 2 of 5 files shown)


pyeudiw/tests/satosa/test_backend.py
```diff
@@ -520,6 +527,150 @@ def test_response_endpoint(self, context):
...
+            "state": state,
+            "vp_token": vp_token,
+            "presentation_submission": {
+                "definition_id": "32f54163-7166-48f1-93d8-ff217bdb0653",
+                "id": "04a98be3-7fb0-4cf5-af9a-31579c8b0e7d",
+                "descriptor_map": [
+                    {
+                        "id": "pid-sd-jwt:unique_id+given_name+family_name",
+                        "path": "$.vp_token.verified_claims.claims._sd[0]",
+                        "format": "vc+sd-jwt"
```
peppelinux, line 597:
```suggestion
                        "format": "dc+sd-jwt"
```

PR #376, opened 2025-03-03
Authorization Error Response Flow

### LINKED ISSUE #368: Support for Authorization Error Response flow
Currenlty the response endpoint supports (allegedly) valid responses containing `state`, `vp_token` and `presentation_submission` (DCQL might change this but the idea is the same).

Hoewever, the response endpoint might receive a POST authorization response with fields `error`, `error_description` and `state` (~eventually wrapped in a jwt if the response mode is `direct_post.jwt`?~)¹.
In this case, the relying party should accept direct_post (error) response and still reply with a 200 Ok response possibly containing a redirect URL that informs the user on what happened (see EDIT below); or at least this what I understand from https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#section-8.2-19

This whole accepting error response flow is currently unimplemented and afaik might also be missing from the stable italian specifications as of 2025-02-24 as I don't see a section "Authorization Error Response Details".

EDIT: this is the content of the page that should be displayed at the user side according to the current (2025-02-25) italian specifications of the openid4vp profile, the RP should responde to its own redirect URL with a 401

¹EDIT 04-03-2025: according to openid4vp draft 24, response encryption are intended to be used for authorization response and this does **NOT** include authorization error response

PascalDR:
Resolves #368

### DIFF
--- a/pyeudiw/satosa/default/response_handler.py
+++ b/pyeudiw/satosa/default/response_handler.py
@@ -92,6 +92,18 @@ class ResponseHandler(ResponseHandlerInterface):
                 f"unable to find nonce in session associated to state {state}: corrupted data"
             )
         return request_session
+    
+    def _handle_error_response(self, error_response: ErrorResponsePayload) -> JsonResponse:
+        request_session: dict = {}
+        request_session = self._retrieve_session_from_state(error_response.state)
+
+        self.db_engine.update_response_object(
+            request_session["nonce"], error_response.state, asdict(error_response), True
+        )
+
+        self.db_engine.set_finalized(request_session["document_id"])
+
+        return JsonResponse({"status": "OK"}, status="200")
 
     def response_endpoint(
         self, context: Context, *args: tuple
@@ -114,16 +126,7 @@ class ResponseHandler(ResponseHandlerInterface):
         )
 
         if isinstance(authz_payload, ErrorResponsePayload):
-            request_session: dict = {}
-            request_session = self._retrieve_session_from_state(authz_payload.state)
-
-            self.db_engine.update_response_object(
-                request_session["nonce"], authz_payload.state, asdict(authz_payload), True
-            )
-
-            self.db_engine.set_finalized(request_session["document_id"])
-
-            return JsonResponse({"status": "OK"}, status="200")
+            return self._handle_error_response(authz_payload)
 
         request_session: dict = {}
         try:
--- a/pyeudiw/satosa/default/response_handler.py
+++ b/pyeudiw/satosa/default/response_handler.py
@@ -3,7 +3,7 @@ import hashlib
 import json
 import logging
 from copy import deepcopy
-from typing import Any
+from typing import Any, Union
 
 from satosa.context import Context
 from satosa.internal import AuthenticationInformation, InternalData
@@ -16,6 +16,7 @@ from pyeudiw.openid4vp.authorization_response import (
     DirectPostParser,
     detect_response_mode,
 )
+from pyeudiw.openid4vp.schemas.response import ErrorResponsePayload
 from pyeudiw.openid4vp.exceptions import (
     AuthRespParsingException,
     AuthRespValidationException,
@@ -98,9 +99,7 @@ class ResponseHandler(ResponseHandlerInterface):
 
         # parse and eventually decrypt jwt in response
         try:
-            authz_payload: AuthorizeResponsePayload = (
-                self._parse_authorization_response(context)
-            )
+            authz_payload = self._parse_authorization_response(context)
         except AuthRespParsingException as e400:
             self._handle_400(context, e400.args[0], e400.args[1])
         except AuthRespValidationException as e401:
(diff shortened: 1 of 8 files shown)


pyeudiw/satosa/default/response_handler.py
```diff
@@ -113,6 +113,18 @@ def response_endpoint(
...
+            request_session: dict = {}
+            request_session = self._retrieve_session_from_state(authz_payload.state)
+
+            self.db_engine.update_response_object(
+                request_session["nonce"], authz_payload.state, asdict(authz_payload), True
+            )
+
+            self.db_engine.set_finalized(request_session["document_id"])
+
+            return JsonResponse({"status": "OK"}, status="200")
```
Zicchio, line 126: The response is not correct and there is a missing half to this PR.

This step records the wallet response in the mongo storage (which is ok) but it not actually elaborating the final user response when the authentication error flow finishes.
In this endpoint we have to return an HTTP 200 OK response that might or might have the redirect URI in the body as [described here](https://italia.github.io/eudi-wallet-it-docs/versione-corrente/en/relying-party-solution.html#redirect-uri). The redirect URI is usually included in the same-device flow - this is the same as regular auth response, so you can look at here for inspiration
https://github.com/italia/eudi-wallet-it-python/blob/0bb0d986b3daf55ae76dc1cf115bc5de6543e85d/pyeudiw/satosa/default/response_handler.py#L199-L204

After that, the wallet or the user-agent (aka the js in polling¹ at the QR code page) will call this redirect uri and the flow will go on this endpoint, who elaborate the "response" to the redirect URI
https://github.com/italia/eudi-wallet-it-python/blob/0bb0d986b3daf55ae76dc1cf115bc5de6543e85d/pyeudiw/satosa/default/openid4vp_backend.py#L313

this endpoint is not currently designed to handle error response for example this dictionary access is broken in case of authorization error responses
https://github.com/italia/eudi-wallet-it-python/blob/0bb0d986b3daf55ae76dc1cf115bc5de6543e85d/pyeudiw/satosa/default/openid4vp_backend.py#L352

Somewhere (probably this endpoint but not 100%) we have to elaborate a final user response (to the redirect_url) to display the proper information based on the wallet authorization error response.
I don't think we have strong indication on what to show the user, it might be up to the intended RP design or target user experience.
If @peppelinux agrees, I would temporarily display a minimal 401 response and eventually open a second issue+PR to investigate what we should actually tell the user in case of **valid** authorization error response.

Other references, which I include because I am not 100% sure about authorization error response and a double check never hurts
https://italia.github.io/eudi-wallet-it-docs/versione-corrente/en/relying-party-solution.html#authorization-error-response
https://italia.github.io/eudi-wallet-it-docs/versione-corrente/en/relying-party-solution.html#redirect-uri-errors
and most importatly
https://italia.github.io/eudi-wallet-it-docs/versione-corrente/en/relying-party-solution.html#device-flow-status-checks-and-security

pyeudiw/storage/base_storage.py
```diff
@@ -117,7 +117,7 @@ def update_request_object(
         raise NotImplementedError()
 
     def update_response_object(
-        self, nonce: str, state: str, response_object: dict
+        self, nonce: str, state: str, response_object: dict, isError: bool = False
```
Zicchio, line 120: remember to also update the docstring when you change a method signature

pyeudiw/tests/satosa/test_backend.py
```diff
@@ -520,6 +520,50 @@ def test_response_endpoint(self, context):
...
+        context.http_headers = {"HTTP_CONTENT_TYPE": "application/x-www-form-urlencoded"}
+
+        response_endpoint = self.backend.response_endpoint(context)
+        msg = json.loads(response_endpoint.message)
+        assert response_endpoint.status == "200"
+
+        doc = self.backend.db_engine.get_by_state(state)
+
+        assert doc["finalized"] == True
+        assert doc["error_response"] == response_with_error
```
Zicchio, line 565: nitpick but we should probably assert that `"error_response"` is a key in the doc before checking it's value, otherwise the test might fail due to test exception rather than a failed assertion (at least it fails anyway, which is the important thing).

pyeudiw/tests/satosa/test_backend.py
```diff
@@ -520,6 +520,50 @@ def test_response_endpoint(self, context):
...
+            "error": "invalid_request",
+            "error_description": "invalid request"
+        }
+
+        context.request = response_with_error
+        context.http_headers = {"HTTP_CONTENT_TYPE": "application/x-www-form-urlencoded"}
+
+        response_endpoint = self.backend.response_endpoint(context)
+        msg = json.loads(response_endpoint.message)
+        assert response_endpoint.status == "200"
```
Zicchio, line 560: since this test is for same device flow, this is missing an assertion that msg (currently unused) contains a non empty redirect_uri claim

see also the big comment above as to why

pyeudiw/satosa/default/response_handler.py
PascalDR: Wow that's a lot of content.
Why all this details are described here and not the issue #368?
I cannot write a PR that cover the 100% of the issue if only the half is disclosed.
Please @Zicchio put all this content in a new issue, omitting the part on 352 that's is clearly part of this PR, where you and @peppelinux can decide on the open points.

peppelinux:
I'd go for merging this and continue working on the issuer later on

peppelinux (approved):
Since this pr brings many changes, the pending review will be handled in a separate PR

Zicchio (approved):

PR #381, opened 2025-03-06
Fix attestation model

### LINKED ISSUE #372: TrustHandler - Cached Public Keys
I don't think [this PR](https://github.com/italia/eudi-wallet-it-python/pull/371/) solves the security issue highlighted here https://github.com/italia/eudi-wallet-it-python/issues/340#issuecomment-2621059431

This is because force_update "pushes" the problem to the trust handler, forcing an update
https://github.com/italia/eudi-wallet-it-python/blob/2dadd24d6ed011e2d3ea8d42825af36b7652870e/pyeudiw/trust/dynamic.py#L76-L78

But then the Trust Handler has no way to know which keys in the TrustSource should be flushed. Stale trust parameters can be updated but old keys in the trust source stay there.
One idea might be to flush *all* keys/trust_parameters from the trust source, then re-evaluate all the trust parameters+keys, but first we should double check if has some unintended side effect. I'm not confident that this solution does not have a bad effect somewhere else.

For what it's worth I think that not solving the problem now is fine. I think that my comment in issue #340 should be de-scoped from this PR and we should eventually open a new issue.

_Originally posted by @Zicchio in https://github.com/italia/eudi-wallet-it-python/issues/371#issuecomment-2689032511_

PascalDR:
Fix #372 

This PR restores the attestation model described in the docs dividing the keys for trust handler.

### DIFF
--- a/example/satosa/integration_test/commons.py
+++ b/example/satosa/integration_test/commons.py
@@ -61,8 +61,12 @@ CREDENTIAL_ISSUER_TRUST_SOURCE_Dict = {
     "policies": {},
     "metadata": {},
     "revoked": False,
-    "keys": [CREDENTIAL_ISSUER_JWK.as_dict()],
-    "trust_params": {}
+    "direct_trust_sd_jwt_vc": {
+        "tp_key": "jwks",
+        "expiration_date": exp_from_now(1024),
+        "jwks": [CREDENTIAL_ISSUER_JWK.as_dict()],
+        "trust_handler_name": "DirectTrustSdJwtVc",
+    },
 }
 CREDENTIAL_ISSUER_TRUST_SOURCE = TrustSourceData(**CREDENTIAL_ISSUER_TRUST_SOURCE_Dict)
 WALLET_PRIVATE_JWK = JWK(leaf_wallet_jwk.serialize(private=True))
--- a/pyeudiw/trust/dynamic.py
+++ b/pyeudiw/trust/dynamic.py
@@ -55,7 +55,12 @@ class CombinedTrustEvaluator(BaseLogger):
         """
         try:
             trust_source = self.db_engine.get_trust_source(issuer)
-            return TrustSourceData.from_dict(trust_source) if trust_source else None
+
+            if trust_source:
+                del trust_source["_id"]
+                return TrustSourceData.from_dict(trust_source)
+
+            return None
         except EntryNotFound:
             return None
         
@@ -97,7 +102,7 @@ class CombinedTrustEvaluator(BaseLogger):
         for handler in self.handlers:
             handler_name = handler.__class__.__name__
 
-            trust_param = trust_source.get_trust_param(handler_name)
+            trust_param = trust_source.get_trust_param_by_handler_name(handler_name)
 
             if not trust_param or trust_param.expired or trust_source.revoked:
                 trust_source = handler.extract_and_update_trust_materials(
@@ -161,13 +166,19 @@ class CombinedTrustEvaluator(BaseLogger):
         """
         trust_source = self._get_trust_source(issuer, force_update)
 
-        if not trust_source.keys:
+        keys = []
+
+        for handler in self.handlers:
+            if (key := trust_source.get_trust_param_by_handler_name(handler.__class__.__name__)) is not None:
+                keys += key.jwks
+
+        if not keys:
             raise NoCriptographicMaterial(
                 f"no trust evaluator can provide cyptographic material "
                 f"for {issuer}: searched among: {self.handlers_names}"
             )
 
-        return trust_source.public_keys
+        return keys
 
     def get_metadata(self, issuer: Optional[str] = None, force_update: bool = False) -> dict:
         """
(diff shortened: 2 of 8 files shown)


docs/STORAGE.md
```diff
@@ -2,65 +2,66 @@
...
+- `get_by_state_and_session_id`
+- `init_session`
+- `set_session_retention_ttl`
+- `has_session_retention_ttl`
+- `add_dpop_proof_and_attestation`
+- `update_request_object`
+- `set_finalized`
+- `update_response_object`
+- `_get_trust_attestation`
+- `get_trust_attestation`
```
peppelinux, line 35: non serve includere i privati se hanno lo stesso nome dei pubblici

sarebbe utile avere una rapida descrizione dello scopo/funzione per metodo

docs/STORAGE.md
```diff
@@ -86,50 +87,50 @@ For a complete list of the MongoDB configuration parameters, see [README.SATOSA.
...
-      },
+      "tp_key": "x5c",
+      "trust_handler_name": "X509Handler",
+      "jwks":  ARRAY[object],
     },
     "direct_trust_sd_jwt_vc": {
-      "jwks": {
-        "keys": ARRAY[object]
-      }
+      "tp_key": "jwks",
```
peppelinux, line 118: tp è l'acronimo di ?

PascalDR: trust parameter key

peppelinux: suggerisco qualcosa di più umano e generico, che ne pensi di attribute_name?

example/satosa/integration_test/commons.py
```diff
@@ -61,8 +61,12 @@
     "policies": {},
     "metadata": {},
     "revoked": False,
-    "keys": [CREDENTIAL_ISSUER_JWK.as_dict()],
-    "trust_params": {}
+    "direct_trust_sd_jwt_vc": {
+        "tp_key": "jwks",
```
Zicchio, line 65: what is `tp_key`? is this `attribute_name`?

pyeudiw/trust/model/trust_source.py
```diff
@@ -40,7 +42,7 @@ def selfissued_jwt_header_trust_parameters(self) -> dict[str, any]:
         :returns: The trust parameters for the self-issued jwt header
         :rtype: dict[str, any]
         """
-        return {self.type: self.trust_params}
+        return {self.type: getattr(self, self.attribute_name)}
```
Zicchio, line 45: Does `self.type` always exists? Maybe I am not understanding correctly but I got the impression that in this PR `type` was dismissed in favour of `attribute_name`

As this method still needed? The proposed solution for jwt header resolution does not invoke this method at all and this method might be unused
https://github.com/italia/eudi-wallet-it-python/blob/49f25463c628ecc47fe2eb271059c4637a2e187e/pyeudiw/trust/dynamic.py#L246

pyeudiw/trust/dynamic.py
```diff
@@ -161,13 +166,23 @@ def get_public_keys(self, issuer: Optional[str] = None, force_update: bool = Fal
...
 
-        if not trust_source.keys:
+        keys = []
+        thumbprints = []
+
+        for handler in self.handlers:
+            if (key := trust_source.get_trust_param_by_handler_name(handler.__class__.__name__)) is not None:
+                for jwk in key.jwks:
+                    if key_from_jwk_dict(jwk) not in thumbprints:
+                        thumbprints.append(key_from_jwk_dict(jwk))
```
peppelinux:
```suggestion
                        thumbprints.append(key_from_jwk_dict(jwk).thumbprint("SHA-256"))
```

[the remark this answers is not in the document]
peppelinux: see my suggestion below.

where the solution can be provided together with the problem statement, we should provide them both to speed up the review.

PascalDR: This is already fixed here: [1a5a7d2](https://github.com/italia/eudi-wallet-it-python/pull/381/commits/1a5a7d2e7f36238ea915ba763934099039e70920)

pyeudiw/trust/dynamic.py
```diff
@@ -240,10 +255,15 @@ def get_jwt_header_trust_parameters(self, issuer: Optional[str] = None, force_up
...
-            param.type: param.trust_params
-            for param in trust_source.trust_params.values()
-        }
+        excluded_fields = ["entity_id", "policies", "metadata", "revoked"]
+
+        headers_params = {}
+
+        for param_name, param_value in trust_source.serialize().items():
+            if param_name not in excluded_fields:
+                headers_params[param_value["attribute_name"]] = param_value[param_value["attribute_name"]]
```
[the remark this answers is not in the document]
peppelinux: the attribute_name is about the chain or where the key is present by value or reference

if kid or trust_chain are used it is important that attribute_name contains the parameter used by name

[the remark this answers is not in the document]
peppelinux: please also note that the attribute_name is neutral about be within the header or the payload.

as jwks is provided within the jwt payload using direct trust, while x5c of trust_chain not within a signed jwt

peppelinux (approved):

pyeudiw/trust/dynamic.py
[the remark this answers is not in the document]
Zicchio: Multiple header claims for the same trust mechanism are not regulated by any standard or draft to date, therefore there is not strong evidence that are needed right now. For this reason, tihs comment will be moved to a dedicated in an issue for future triage.

PR #406, opened 2025-04-09
Check that the first key is properly configured

### LINKED ISSUE #397: Consider supported alg in the presentation request JWT
This was closed but note that PR #394 does NOT enforce the usage of `config.jwt.default_sig_alg`, `config.jwt.default_enc_alg` and `config.jwt.default_enc_enc` when the RP sign/encrypt token; for example when it produces a JAR here
> https://github.com/italia/eudi-wallet-it-python/blob/1c02a26bd903a57034d987a3d62c4bf9db5efffb/pyeudiw/satosa/default/request_handler.py#L66-L86
> 
> in a similar way, `config.jwt.enc_alg_supported` and `config.jwt.enc_enc_supported` are used when validating a direct_post.jwt token response
> https://github.com/italia/eudi-wallet-it-python/blob/523630277b7fccc9d0d5af2010d3d226ef8b0c36/pyeudiw/openid4vp/authorization_response.py
> but not, say, when validating an sd-jwt with verifiable credentials.
> 
> This suggestion
>These parameter MUST be properly handled in the JWS Helper.
> 
> is probably the best approach in enforcing a solution that is automatically valid at project-level. 

 _Originally posted by @Zicchio in [#395](https://github.com/italia/eudi-wallet-it-python/issues/395#issuecomment-2766143326)_

PascalDR:
Fix #397

### DIFF
--- a/example/satosa/pyeudiw_backend.yaml
+++ b/example/satosa/pyeudiw_backend.yaml
@@ -33,7 +33,7 @@ config:
     sym_key: "1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdef" # hex string of 64 characters
 
   jwt:
-    default_sig_alg: RS256 # or ES256. Please note that this signature alg MUST be compliant with the private keys used for the signature. X.509 certificates MUST be therefore ECDSA using ES, and RSA using RS
+    default_sig_alg: ES256 # or RS256. Please note that this signature alg MUST be compliant with the private keys used for the signature. X.509 certificates MUST be therefore ECDSA using ES, and RSA using RS
     default_enc_alg: RSA-OAEP
     default_enc_enc: A256CBC-HS512
     default_exp: 6         # minutes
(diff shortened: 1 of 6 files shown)


example/satosa/pyeudiw_backend.yaml
```diff
@@ -33,7 +33,7 @@ config:
     sym_key: "1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdef" # hex string of 64 characters
 
   jwt:
-    default_sig_alg: ES256 # or RS256. Please note that this signature alg MUST be compliant with the private keys used for the signature. X.509 certificates MUST be therefore ECDSA using ES, and RSA using RS
+    default_sig_alg: RS256 # or ES256. Please note that this signature alg MUST be compliant with the private keys used for the signature. X.509 certificates MUST be therefore ECDSA using ES, and RSA using RS
```
peppelinux, line 36: default sign alg must be ES256

Issue #408, opened 2025-04-10
Integration tests stop working - TrustSource inizialization inconsistencies


Zicchio:
This is probably already known, but it looks like end-to-end test locate here
https://github.com/italia/eudi-wallet-it-python/tree/main/example/satosa/integration_test
are no longer up to date and should be updated to reflect and validate the changes from e5fef59a412a3caf0bad98f0b60ddfce3f2a4638 up to ced6fe3ae652dabcd9444488305059ffbca6cd90

peppelinux:
the issue is this

````
[2025-04-10 16:46:34] [ERROR]: [urn:uuid:57954884-24aa-427e-bf98-6d6cd414e802] {'message': 'Uncaught exception', 'error': 'pyeudiw.trust.model.trust_source.TrustEvaluationType() argument after ** must be a mapping, not list', 'error_id': 'urn:uuid:85a4d4a4-99cb-4ccb-9c73-882a0ab16e6a'} [satosa.base.run:375]
````

the inizialization of the Trust Source seems inconsistent ([here](https://github.com/italia/eudi-wallet-it-python/blob/main/pyeudiw/trust/model/trust_source.py#L113) the mapping requires a json object)

`./trust/model/trust_source.py:112:            setattr(self, _type, TrustEvaluationType(**tp))`

it turns out the in the DB we have a list containing multiple json objects each of them representing a jwk

![Image](https://github.com/user-attachments/assets/19fbd39c-80cb-4870-abb4-b14d2d82b633)

the point is that we have misalignments between the unit test configuration and what actually happend within the integration tests

peppelinux:
integration tests must be executed for any opened PR

peppelinux:
The issue not just about the `keys` data schema, but also about this

````
TypeError: TrustEvaluationType.__init__() missing 2 required positional arguments: 'attribute_name' and 'expiration_date'
````

Issue #378, opened 2025-03-04
Authorization Error Response: technical implementation details


Zicchio:
The purpose of this issue is to expand on the topic of issue #368, which contains a brief description of a missing feature without including a full suggestion on how to implement it.

To contextualize the changes required by a full resolution of issue #368 current authentication flows follows the  diagram described in [the italian specs](https://italia.github.io/eudi-wallet-it-docs/versione-corrente/en/relying-party-solution.html), which can be summarized as:

1. a pre authentication / login begin step, where either a QR code is displayer (cross device flow) or the wallet app is inoked via deep link in the user smartphone; the http handler associated to this behaviour is this
https://github.com/italia/eudi-wallet-it-python/blob/0bb0d986b3daf55ae76dc1cf115bc5de6543e85d/pyeudiw/satosa/default/openid4vp_backend.py#L219
1.1. in case of cross device flow, a javascript is polling on the /status endpoint; the primary poropose of this polling is to notify when step 3. is finished. https://github.com/italia/eudi-wallet-it-python/blob/0bb0d986b3daf55ae76dc1cf115bc5de6543e85d/pyeudiw/satosa/default/openid4vp_backend.py#L356
2. a "jar" request step: the wallet calls via HTTP request the url in the QR code (or in the deep link) to obtain the authentication request jwt that includes, in the claims, the oauth request performed by the RP toward the Wallet; the http handler associated to this behaviour is this
https://github.com/italia/eudi-wallet-it-python/blob/0bb0d986b3daf55ae76dc1cf115bc5de6543e85d/pyeudiw/satosa/default/openid4vp_backend.py#L219
3. an authorizatin response step: the wallet sends the to the wallet the user digital credential via an HTTP; the handler associated to that endpoint is this https://github.com/italia/eudi-wallet-it-python/blob/0bb0d986b3daf55ae76dc1cf115bc5de6543e85d/pyeudiw/satosa/default/response_handler.py#L94
the response to this endpoint is either
3.1 [Cross Device Flow] 200 Ok response; then the /status endpoint will finally get an ok response containing a redirect URL in its body (that is invoked by the user agent) to go on with step 4.
3.2 [Same Device Flow] 200 Ok with a redirect url to be invoked at step 4.
[...]

PR #411, opened 2025-04-11
feat: authorization error response

Zicchio:
This PR is **PARTIAL** attempt at dealing with #378. The goal of this PR is to provide a **minimal** working error error management for cases when the user denies the authentication request.

Implementation details as follows in scenarios where the wallet sends an authorization error response.
1. The status endpoint was updated to return 401 errors. As a consequence, the page with a QR code and polling javascript will update accordingly (relevant for cross device flow).
2. The redirect_uri leads to a generic error page¹ signaling that the authentication was canceled or failed (relevant for same device flow).
3. Suggestions proposed here https://github.com/italia/eudi-wallet-it-python/issues/378#issuecomment-2790346402 were not taken into account

A new integration test was added to test this case, as such **THIS PR MIGHT GENERATE CONFLICT WITH SOMEONE DEALING WITH #408 **.

¹ It is ugly and not very usable, and should be treated as a proof of concept.

### DIFF
--- a/example/satosa/integration_test/same_device_integration_test_duckle.py
+++ b/example/satosa/integration_test/same_device_integration_test_duckle.py
@@ -1,49 +1,4 @@
 # This file defines an end-to-end integration test flow for Duckle support.
-#
-# To run this integration test, you need to modify the `pyeudiw_backend.yaml` configuration file
-# by adding the following entries:
-#
-# config:
-#   duckle:
-#     dcql_query: >
-#       {
-#         "credentials": [
-#           {
-#             "id": "personal id data",
-#             "format": "dc+sd-jwt",
-#             "meta": {
-#               "vct_values": [
-#                 "https://trust-registry.eid-wallet.example.it/credentials/v1.0/personidentificationdata"
-#               ]
-#             },
-#             "claims": [
-#               { "path": ["given_name"] },
-#               { "path": ["family_name"] }
-#             ]
-#           },
-#           {
-#             "id": "wallet attestation",
-#             "format": "dc+sd-jwt",
-#             "meta": {
-#               "vct_values": [
-#                 "https://itwallet.registry.example.it/WalletAttestation"
-#               ]
-#             },
-#             "claims": [
-#               { "path": ["wallet_link"] },
-#               { "path": ["wallet_name"] }
-#             ]
-#           }
-#         ]
-#       }
-#
-# Additionally, you must register the Duckle handler in the `credential_presentation_handlers` section:
-#
-# credential_presentation_handlers: [...]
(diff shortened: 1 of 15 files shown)


example/satosa/pyeudiw_backend.yaml
```diff
@@ -101,9 +101,6 @@ config:
             - path:
               - "$.given_name"
 
-  duckle:
```
peppelinux, line 103: why it is removed?

example/satosa/integration_test/testconfig/potential/wp2uc1/userdenies/pyeudiw_backend.yaml
```diff
@@ -194,8 +195,6 @@ config:
       - module: pyeudiw.openid4vp.vp_sd_jwt_vc
         class:  VpVcSdJwtParserVerifier
         format: dc+sd-jwt
-        config:
```
peppelinux: why it is removed?

example/satosa/integration_test/same_device_integration_test_duckle.py
```diff
@@ -1,3 +1,50 @@
+# This file defines an end-to-end integration test flow for Duckle support.
+#
+# To run this integration test, you need to modify the `pyeudiw_backend.yaml` configuration file
```
peppelinux, line 3: no, duckle is now the default query language, therefore the example configuration must have duckle by default

peppelinux (changes requested):

example/satosa/pyeudiw_backend.yaml
saralongobardiacn: Restored 
https://github.com/italia/eudi-wallet-it-python/pull/411/commits/049cde8d3bc3b050d51adbdf7f4df27c6a994614

example/satosa/integration_test/same_device_integration_test_duckle.py
saralongobardiacn: Restored 
https://github.com/italia/eudi-wallet-it-python/pull/411/commits/049cde8d3bc3b050d51adbdf7f4df27c6a994614

example/satosa/integration_test/testconfig/potential/wp2uc1/userdenies/pyeudiw_backend.yaml
saralongobardiacn: During the handler initialization, the sig_alg_supported configurations are retrieved from the global configuration `config.jwt.sig_alg_supported`, in order to reduce the number of occurrences within the configuration itself.

peppelinux (approved):

Issue #422, opened 2025-04-23
Header x5c in JWT is not compliant with RFC 7517


Zicchio:
When x509 trust is configured, header x5c includes the trust chain in PEM format as configured in file https://github.com/italia/eudi-wallet-it-python/blob/e5a1cb7f0e38498f4c7f509d5c522713fc063c2b/example/satosa/pyeudiw_backend.yaml#L265-L306

However according to RFC 7517 the header x5c should be Base64 encoding of the DER encoding (that is, without the `-----BEGIN CERTIFICATE-----` header and no newlines).

Zicchio:
This issue also appears the other way around.
When reading a token header `x5c` claim, tries to parse the header as PEM rather than Bare64+DER as instructed by RFC 7517

Zicchio:
I think I might be going crazy over this.

Certificates are used in both JWT and COSE (associated to mdoc cbor) headers, so both problems should be considered when designing a solution.

In case JWT, RFC 7517 (and 7515) leaves no ambiguity: header `x5c` is the Base64 encoding of the DER (binary) certificate (chain).
When accepting the static trust material of a JWT x5c header, this assumption is sound and strong.

In case of COSE, things are not clear to me.
First of all, COSE does NOT use x5c, but x5chain, see this section https://www.rfc-editor.org/rfc/rfc9360.html#name-x509-cose-header-parameters
Then, COSE just states on x5chain: "This header parameter contains an ordered array of X.509 certificates". No representation format is suggested.
So how are certificates represented in a a COSE x5chain header. Is it PEM? Is it Base64+DER? Is it DER? Is it ANS.1?
Who knows. Examples are crazy hard to find.
If somebody would guide me to an answer, it would be greatly appreciated.

Finally, I want to quote a phrase from David Wong, _Real World Cryptography_: “I don't recommend any modern application to use X.509 certificates unless it has to”. After all, [validating SSL certificates is the most dangerous code in the world](https://crypto.stanford.edu/~dabo/pubs/abstracts/ssl-client-bugs.html)...

Zicchio:
Ultimately here https://github.com/Zicchio/eudi-wallet-it-python/commit/c7f18e9c0c2c8efb618a16bf1c07dd1370b26926 I decided to handle the problem as follow:
- when validating static trust material ()which might originate from x5c SD-JWT or x5chain MDOC), try to guess the representation format and normalize to DER - this should be trated as **very unsafe**.
[...]

Issue #416, opened 2025-04-16
Missing documentation on how to configure x509 trust handler


Zicchio:
This [document](https://github.com/italia/eudi-wallet-it-python/blob/27b149381fa7b68a201c86540a8841d7e5d83af7/docs/TRUST.md) is missing a section on how to configure an x509 trust hanlder and how each configuration element of such a trust handler is intended to be filled.

For example, in the example configuration

https://github.com/italia/eudi-wallet-it-python/blob/27b149381fa7b68a201c86540a8841d7e5d83af7/example/satosa/pyeudiw_backend.yaml#L251-L264

it is not clear to me what each entry is intended to be. I get that the high level goal is to provide a list of certificate authorities and their (trusted) certificates, but to me it is not clear what the key of each entry (like `ca.example.com`) is actually intended to be. 

For example, say that I want to add this entry as certificate authority (courtesy of our finnish parter)

```
-----BEGIN CERTIFICATE-----
MIIC6jCCAm+gAwIBAgIURxcFSmf0DJh/423u13fx2l9/7pgwCgYIKoZIzj0EAwIw
gagxODA2BgNVBAMML0RWViBEaWdpLUlEIE1vY2sgQXR0cmlidXRlIFNlYWxpbmcg
Q2VydGlmaWNhdGVzMSYwJAYDVQQKDB1EaWdpLSBqYSB2w6Rlc3TDtnRpZXRvdmly
YXN0bzERMA8GA1UEBwwISGVsc2lua2kxCzAJBgNVBAYTAkZJMRAwDgYDVQQIDAdG
aW5sYW5kMRIwEAYDVQQFEwkwMjQ1NDM3LTIwHhcNMjQwNTE2MDg0NzQxWhcNMjUw
NTE2MDg0NzQxWjCBpzE3MDUGA1UEAwwuRFZWIERpZ2ktSUQgTW9jayBBdHRyaWJ1
dGUgU2VhbGluZyBDZXJ0aWZpY2F0ZTEmMCQGA1UECgwdRGlnaS0gamEgdsOkZXN0
w7Z0aWV0b3ZpcmFzdG8xETAPBgNVBAcMCEhlbHNpbmtpMQswCQYDVQQGEwJGSTEQ
MA4GA1UECAwHRmlubGFuZDESMBAGA1UEBRMJMDI0NTQzNy0yMFkwEwYHKoZIzj0C
[...]

PR #431, opened 2025-04-28
Update x509 Trust Docs

Zicchio:
This PR cloes #416 by specifying that x509 chains should be configured in PEM format only.

This neccessity emerged after #428 and #422 highlited that having certificates in uncertain format made some code paths unstable.


Zicchio:
@peppelinux the pipeline failed for some unit test but this PR did not include any code changes.

Probably something went wrong in other merges.

This commit should trivially fix the failing test
https://github.com/italia/eudi-wallet-it-python/pull/411/commits/351cfd039bde8fef4898c675053c2075638b5f51
but PR 411 SHOULD NOT be accepted as of now.

I will isolate this commit to a new branch. so that this PR can be reviewed.

PR #437, opened 2025-04-30
feat: refactor of jwt key selection for x5c

### LINKED ISSUE #423: Extend/move x5c key selection when signing JWT from http handler to JWS helper
This block of code in `request_handler.py` is dedicated to figuiring out which key should be used to sign a Request Object JWT based on the `x5c` header and to validate that a chain actually refers to the signing key.
https://github.com/italia/eudi-wallet-it-python/blob/e5a1cb7f0e38498f4c7f509d5c522713fc063c2b/pyeudiw/satosa/default/request_handler.py#L77-L92

However, there is already a general purpose method dedicated to figuring out which key should be used to sign a token based on the current token header, which is this one and is used by the JWS Helper.

This is used for key selection based on the value of a JWT header
https://github.com/italia/eudi-wallet-it-python/blob/e5a1cb7f0e38498f4c7f509d5c522713fc063c2b/pyeudiw/jwt/jws_helper.py#L159-L185

And this is a sanity check that the selected key is coeherent with the current JWT header
https://github.com/italia/eudi-wallet-it-python/blob/e5a1cb7f0e38498f4c7f509d5c522713fc063c2b/pyeudiw/jwt/jws_helper.py#L109-L115

As such, it would be best to extend~update the JWS helper/signer responsability to include the selection+check concerning x5c header claim done in the http handler `request_hanlder.py`

Zicchio:
This PR closes #423 and in particular does the following:
1. when signing a jwt with a set of available keys and header x5c is in the token, use a key that matches the x5c header (refactored from `request_handler.py` to `jws_helper.JWSHelper`)
2. when signing a token, a consistency double check is performed to verify that the signing key indeed matches possible "kid" header and "x5c" header (when present in both the key and the header).
3. some minor fixes here and there discovered when working on issue 423.

@peppelinux commits can be squashed

### DIFF
--- a/pyeudiw/jwt/jws_helper.py
+++ b/pyeudiw/jwt/jws_helper.py
@@ -10,6 +10,7 @@ from cryptojwt.jwk.jwk import key_from_jwk_dict
 from pyeudiw.jwk import JWK
 from pyeudiw.jwk.exceptions import KidError
 from pyeudiw.jwk.jwks import find_jwk_by_kid, find_jwk_by_thumbprint
+from pyeudiw.jwk.parse import parse_b64der
 from pyeudiw.jwt.exceptions import (
     JWEEncryptionError,
     JWSSigningError,
@@ -76,9 +77,12 @@ class JWSHelper(JWHelperInterface):
         of available keys.
 
         If the header already contains indication of a key, such as 'kid',
-        'trust_chain' and 'x5c', there is no guarantee that the signing
-        key to be used will be aligned with those header. We assume that is
-        it responsibility of the class initiator to make those checks.
+        'trust_chain' and 'x5c', the method will attempt to match the signing
+        key among the available keys based on such claims, but there is no
+        guarantee that the correct key will be selected. We assume that is
+        it responsibility of the class initiator to make those checks. To
+        avoid any possible ambiguity, it is suggested to initilize the class
+        with one (signing) key only.
 
         :param plain_dict: The payload to be signed.
         :param protected: Protected header for the JWS.
@@ -106,13 +110,10 @@ class JWSHelper(JWHelperInterface):
         if signing_key["kty"] == "oct":
             raise JWSSigningError(f"Key {signing_key['kid']} is a symmetric key")
 
-        # Ensure the key ID in the header matches the signing key
-        header_kid = protected.get("kid")
-        signer_kid = signing_key.get("kid")
-        if header_kid and signer_kid and (header_kid != signer_kid):
-            raise JWSSigningError(
-                f"token header contains a kid {header_kid} that does not match the signing key kid {signer_kid}"
-            )
+        try:
+            _validate_key_with_jws_header(signing_key, protected, unprotected)
+        except Exception as e:
+            raise JWSSigningError(f"failed to validate signing key: it's content it not valid for current header claims: {e}", e)
 
         payload = serialize_payload(plain_dict)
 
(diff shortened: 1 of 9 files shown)


pyeudiw/jwt/jws_helper.py
```diff
@@ -210,6 +216,19 @@ def _select_key_by_kid(self, headers: tuple[dict, dict]) -> dict | None:
             return None
         return find_jwk_by_kid([key.to_dict() for key in self.jwks], kid)
 
+    def _select_key_by_x5c(self, headers: tuple[dict, dict]) -> dict | None:
+        if not headers:
+            return None
+        x5c: list[str] | None = headers[0].get("x5c") or headers[1].get("x5c")
```
PascalDR, line 222: Probably the type of headers is list[tuple[str, list[str]]]

pyeudiw/x509/verify.py
```diff
@@ -149,22 +164,28 @@ def to_PEM_cert(cer: str | bytes) -> str:
...
-    else:
-        return DER_cert_to_PEM_cert(cert_s)
+    try:
+        cert_s = cert_b.decode()
+        if _BASE64_RE.fullmatch(cert_s):
+            return B64DER_cert_to_PEM_cert(cert_s)
+    except UnicodeError:
+        return DER_cert_to_PEM_cert(cert_b)
+
+    raise ValueError("unable to recognize input [cert] as a certificate")
```
PascalDR, line 188:
```suggestion
    raise ValueError(f"unable to recognize input {cert} as a certificate")
```

PascalDR:
Some unit tests fails with PyMdocCbor set to 8.0 but all tests pass with 9.0.
@peppelinux a note in the changelog could be useful for the update process.

pyeudiw/jwt/jws_helper.py
Zicchio: Care to explain why?
This is the x5c claims as dicatated here https://datatracker.ietf.org/doc/html/rfc7517#section-4.7
Maybe I am missing something

pyeudiw/x509/verify.py
Zicchio: I'm ok with the change but note that `[cert]` was intentional to not bloat an error message with something that is likely to be unrecognizable.

Zicchio:
> Some unit tests fails with PyMdocCbor set to 8.0 but all tests pass with 9.0. @peppelinux a note in the changelog could be useful for the update process.

100% agree: pymdoc version 0.9 became a requirement some time ago before this pull request and that change was not tracked anywhere not is flagged in the requirements file or the setup.py requirements entry.

I think this problem is worty of opening an issue as in the current sup-up different devs might have a problem in recreating the same environment (typical python issue...)

pyeudiw/jwt/jws_helper.py
Zicchio: Ah, you mean the hint in the fuction signature.

Ye you are right, it should be something like `dict[str, Any]` which is a relaxed type hint for something that resembles an arbitrary JSON object

PascalDR: Nevermind, i confused tuple with dict reading the code.

pyeudiw/x509/verify.py
PascalDR:
```suggestion
    raise ValueError("unable to recognize input as a certificate")
```

Zicchio: accepted in [e5ab0ba](https://github.com/italia/eudi-wallet-it-python/pull/437/commits/e5ab0babf56f7b6e6171979059af41c7fdc48d17)

PascalDR: The function takes just one parameter as input so it's implicit.

PascalDR (approved):

peppelinux:
bump pymdoccbor here https://github.com/italia/eudi-wallet-it-python/commit/1f7f3a5955fce4ca0359a5eba6b83288917fb3be

PR #447, opened 2025-05-28
chore: small improvements in x509 example chain build, unit test

peppelinux:
this PR has introduces few changes improving the gen_chain unit test function for the purpose to use it as example builder for the eid-wallet-it-docs


PascalDR:
@peppelinux 
The error is a bad validation of the chain for the error: x509 verification failed: permitted subtree violation

Using rules like 'x509.DNSName("example.com")' in exclude_subtrees automatically exclude all kind of dns that includes "example.com".
For example "sub.example.com" or "www.example.com" are automatically excluded.

More info:
http://web.archive.org/web/20190719142044/https://www.sysadmins.lv/blog-en/x509-name-constraints-certificate-extension-all-you-should-know.aspx

PR #485, opened 2025-07-29
[Interoperability] Oauth2 par request

PascalDR:
This PR removes the work done in #484 in favor on an approach more dynamic.
As described in in the PAR [section in Openid4vci](https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#name-pushed-authorization-reques) in the body of the request are included only the fields are: scope, response_type, client_id, code_challenge, code_challenge_method, state, redirect_uri.
For this reason it can't be mapped 1:1 with an IT Wallet request because the signed jwt has more fields that are validated like exp or jti.
To enable the interoperability between IT Wallet and the classical Openid4VCI the PAR endpoints checks that the body contains a request parameter.
If present, and the signed_par_request is true or both, it validate and decode the body of JWT and does all extra necessary validation on the fields, omitting the wallet attestation or the dpop validations if the relative fields in security section in the config are set to false, if not present it will be validated as a classical Openid4VCI request.

### DIFF
--- a/pyeudiw/satosa/frontends/openid4vci/endpoints/authorization_endpoint.py
+++ b/pyeudiw/satosa/frontends/openid4vci/endpoints/authorization_endpoint.py
@@ -14,7 +14,7 @@ from pyeudiw.satosa.frontends.openid4vci.models.openid4vci_basemodel import (
     ENDPOINT_CTX,
     CLIENT_ID_CTX,
 )
-from pyeudiw.satosa.frontends.openid4vci.storage.engine import OpenId4VciEngine
+from pyeudiw.satosa.frontends.openid4vci.storage.engine import OpenId4VciDBEngineHandler
 from pyeudiw.satosa.frontends.openid4vci.tools.exceptions import InvalidRequestException
 from pyeudiw.satosa.utils.session import get_session_id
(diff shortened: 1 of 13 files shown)


peppelinux:
@PascalDR from your analysis it seems that it-wallet forces the signature of the request, while openidvci does not.

the approach using an unsigned request seems to be compliant to openid haip as well: https://openid.github.io/oid4vc-haip/openid4vc-high-assurance-interoperability-profile-wg-draft.html#section-4.3

it seems that there is not explicit mention to signed request objects in openid4vci, while rfc9126 does (https://datatracker.ietf.org/doc/html/rfc9126) and it is referenced by openid4vci. using signed request objects, exp and aud parameters must be used and checked as well.

the argument is more about the interop success linked to the popularity of using signed PAR in the wallet ecosystem.

at the current stage, thank to your contribution, eudiw-wallet-it-python and its satosa frontend now fully supports both approaches.

peppelinux (approved):

peppelinux:
@PascalDR please, the deployers must be able to decide of the signatiue of the request Is required or optional, there the security configuration parametri allow_unsigned_par_request should be implemented

PR #488, opened 2025-08-05
Implementation of wallet header proofs

### LINKED ISSUE #486: Deployers must be able to decide of the signatue of the request Is required or optional
@PascalDR please, the deployers must be able to decide of the signatiue of the request Is required or optional, there the security configuration parametri allow_unsigned_par_request should be implemented

_Originally posted by @peppelinux in https://github.com/italia/eudi-wallet-it-python/issues/485#issuecomment-3141687516_

PascalDR:
Resolve #486


peppelinux:
we have a huge semantical misunderstanting here

DPOP is only about Access Token, see: https://datatracker.ietf.org/doc/html/rfc9449

Client Authz uses its own proof of possession mechanism, see: https://www.ietf.org/archive/id/draft-ietf-oauth-attestation-based-client-auth-06.html#section-5.2

this means that security.dpop_required MUST only pertaing AS's token endpoint and credential endpoint

while security.wallet_attestation_required implicltly requires its jwt pop

Please, align both the code and the documentation with the drafts and standards defining dpop for access token only, and jwt pop for the client attestation as default behaviour when client attestation at PAR endpoint is enabled

docs/OPENID4VCI-SATOSA-FRONTEND.md
```diff
@@ -285,44 +285,46 @@ Metadata for OpenID Federation.
...
+| config.security.dpop_required               | This is the flag to enable the DPOP validation needed as access token for token and credential endpoints<br />([https://datatracker.ietf.org/doc/html/rfc9449](https://datatracker.ietf.org/doc/html/rfc9449)).<br /> If this setting is missing it will be set to true by default.                                                                                       |
+| config.security.wallet_attestation_required | This is the flag to enable the wallet client attestation validation in POP JWT format<br />([https://www.ietf.org/archive/id/draft-ietf-oauth-attestation-based-client-auth-06.html#section-5.2](https://www.ietf.org/archive/id/draft-ietf-oauth-attestation-based-client-auth-06.html#section-5.2)).<br />If this setting is missing it will be set to true by default. |
```
peppelinux, line 303:
```suggestion
| config.security.wallet_attestation_required | This flag enables the [OAuth 2.0 Attestation-Based Client Authentication](https://www.ietf.org/archive/id/draft-ietf-oauth-attestation-based-client-auth-06.html). If this setting is missing it will be set to true by default. |
```

pyeudiw/satosa/utils/validation.py
```diff
@@ -52,7 +52,24 @@ def validate_request_method(request_method: str, accepted_methods: list[str]):
...
+    """
+    Validates the presence of the DPoP proof in the request headers.
+    Args:
+        context (Context): The SATOSA context containing the HTTP request.
+    Raises:
+        InvalidRequestException: If the DPoP proof header is missing.
+    """
+    header_pop = context.http_headers.get(OAUTH_CLIENT_ATTESTATION_POP_HEADER)
+
+    #TODO: add dpop validation
```
peppelinux, line 66: why TODO?

peppelinux: this must validate the oauth client attestation pop header and payload

PR #490, opened 2025-09-08
fix: RemoteFlowType.CROSS_DEVICE is always detected

### LINKED ISSUE #405: RemoteFlowType.CROSS_DEVICE is always detected
I am working on a credential issuing flow with PID authentication and ran into that the RemoteFlowType.CROSS_DEVICE is always detected. 
We want the SAME_DEVICE flow so I am hard-coding it for now.

saralongobardiacn:
**It needs to be tested with wwallet, so we won’t merge it before first doing a test with the deployment.**

Fix #405


integration_test/conf/pyeudiw_backend.yaml
```diff
@@ -344,6 +344,14 @@ config:
 
         private_keys: *metadata_jwks
 
+  security:
+    accepted_http_referers: # list of referers that indicate a wallet-originated request. Requests from these URLs will use SAME_DEVICE flow even on desktop.
```
peppelinux, line 348: this would end up with an infinite list of wallet solution

I would not go in this direction of whitelisting

saralongobardiacn: A change has been made to accept as input an optional list of regex patterns matched against the HTTP_REFERER header. If the list is empty or the key is missing, no referer-based enforcement is applied and all referers are accepted. If any regex matches, the flow is forced to SAME_DEVICE.
View commit: [fix: RemoteFlowType.CROSS_DEVICE is always detected, support SAME_DEVICE flow based on trusted referers regex criteria](https://github.com/italia/eudi-wallet-it-python/pull/490/commits/e353ecf5ce2c19a44cac5b58d75d98e91ec2d1d4)

docs/OPENID4VP-SATOSA-BACKEND.md
```diff
@@ -177,6 +177,14 @@ For more deatils on available trust implementations and their configurations, se
...
+|-------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| config.security.accepted_referer_criteria | An optional list of regular expressions that are matched against the HTTP_REFERER header of an incoming request. <br/> If the referer matches one of these regex patterns, the system will force the same-device flow (SAME_DEVICE) even if the client is on a desktop or non-mobile device. <br/> This enables use cases where a user already has a credential in their wallet and is scanning a QR code to obtain a new credential, allowing authentication using the existing PID without requiring a cross-device flow. |
```
peppelinux, line 186:
```suggestion
| config.force_same_device_flow_referer_criteria | An optional list of regular expressions that are matched against the HTTP_REFERER header of an incoming request. <br/> If the referer matches one of these regex patterns, the system will force the same-device flow (SAME_DEVICE) even if the client is on a desktop or non-mobile device. <br/> This enables use cases where a user already has a credential in their wallet and is scanning a QR code to obtain a new credential, allowing authentication using the existing PID without requiring a cross-device flow. |
```

I would put it swithin the security context, since its purpose is about interop only and same device flow

therefore also the name of the parameter must be more eloquent, such as the one suggested above

peppelinux (changes requested):

docs/OPENID4VP-SATOSA-BACKEND.md
saralongobardiacn: View commit:
[fix: RemoteFlowType.CROSS_DEVICE is always detected, rename properties for forcing](https://github.com/italia/eudi-wallet-it-python/pull/490/commits/fcd9dbd7ae1ca2a670eebb569979fb88f6684804)

docs/OPENID4VP-SATOSA-BACKEND.md
```diff
@@ -177,6 +177,14 @@ For more deatils on available trust implementations and their configurations, se
...
+|------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| config.force_same_device_flow_referer_criteria | An optional list of regular expressions that are matched against the HTTP_REFERER header of an incoming request. <br/> If the referer matches one of these regex patterns, the system will force the same-device flow (SAME_DEVICE) even if the client is on a desktop or non-mobile device. <br/> This enables use cases where a user already has a credential in their wallet and is scanning a QR code to obtain a new credential, allowing authentication using the existing PID without requiring a cross-device flow. |
```
peppelinux, line 186:
```suggestion
| config.force_same_device_flow_referer_criteria | An optional list of regular expressions that are matched against the HTTP_REFERER header of an incoming request. <br/> If the referer matches one of these regex patterns, the system will force the same-device flow (SAME_DEVICE) even if the client is on a desktop or non-mobile device. <br/> This enables use cases where a wallet instance is provided as Single Page Application (eg: wwWallet). |
```

PR #499, opened 2025-09-22
X509 chain fix3

peppelinux:
This PR further improves the example X509 chain used in the unit tests and according to eid-wallet-it-docs and further ETSI requirements


pyeudiw/x509/verify.py
```diff
@@ -252,7 +252,7 @@ def verify_x509_anchor(pem_str: str) -> bool:
     """
     cert_data = load_der_x509_certificate(to_DER_cert(pem_str))
 
-    if not _check_datetime(cert_data.not_valid_after):
+    if not _check_datetime(cert_data.not_valid_after_utc):
```
peppelinux, line 255: this removes the warning shown during the execution of the tests

PR #502, opened 2025-09-26
Revert "Implemented jwt vc issuer metadata endpoint handler"

peppelinux:
Reverts italia/eudi-wallet-it-python#500


integration_test/conf/openid4vci_frontend.yaml
```diff
@@ -52,10 +52,6 @@ config:
       module: pyeudiw.satosa.frontends.openid4vci.endpoints.oauth_authorization_server_metadata_endpoint
       class: OauthAuthorizationServerMetadataHandler
       path: '.well-known/oauth-authorization-server'
-    jwt_vc_issuer_metadata:
```
peppelinux, line 55: @PascalDR here as default endpoint had sense

PR #528, opened 2026-02-17
Fix Private Key Exposure in JWKS

### LINKED ISSUE #526: [OpenID4VCIFrontend] Security Vulnerability - Private Key Exposure in JWKS
In OpenID4VCIFrontend the JWKS (JSON Web Key Set) provided in Entity Configuration (`/.well-known/openid-federation`), includes private key parameters (e.g. d).
According to the [OpenID4VCI specification](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html), the JWKS must only expose public keys. These keys should never include sensitive private material.

**Impact:**
Exposing the private key completely breaks the trust model of the identity ecosystem.

**Metadata example:**
```
{
  "iss": "https://eaa-provider.example.org",
  "sub": "https://eaa-provider.example.org",
  "jwks": {
    "keys": [
      {
        "kty": "EC",
        "d": "", // <--- ERROR: private element
        "use": "",
        "crv": "P-256",
        "kid": "",
        "x": "",
        "y": "",
        "alg": "ES256"
      }
    ]
  }
}
```

manpace:
fixed issue #526 

 Changelog:
 - removed private jwks from EC
 - added federation jwks to EC


peppelinux (approved):

pyeudiw/satosa/frontends/openid4vci/endpoints/metadata_endpoint.py
```diff
@@ -34,14 +35,20 @@ def _ensure_credential_issuer(self, metadata: dict, metadata_key: str, issuer_ke
     @property
     def metadata(self) -> dict:
         metadata = self.config.get("metadata", {})
-        [
```
peppelinux, line 37: @manpace please use the list comprension as made before, just by adding the public key only

the code you provided overloads the same sinlge object multiple times, whjile the list ompresion not

peppelinux (changes requested):

peppelinux (approved):

PR #531, opened 2026-03-13
wallet attestations

manpace:
wallet app attestation
wallet app attestation request
wallet unit attestation


pyeudiw/tests/wallet_instance_attestations/test_attestation_request.py
peppelinux, line 1: this lib must be renamed in a more general way

wallet_attestations instead of wallet_instance_attestation

peppelinux:
@manpace the CI is broken and the componest is still not renamed to wallet_attestations after more than 2 weeks

doing it by myself, I would ask you when possible the reasons about these performance

pyeudiw/tests/wallet_instance_attestations/test_attestation_request.py
peppelinux: done in 7165263b831e797094e7531020a06861132c233c

peppelinux:
@manpace the content of this PR was still misaligned with both eudiw ARF and it-wallet specs

- x5c is mandatory in wia jtw header
- the typ of both wia and wa request was misaligned
- the sign alg cannot be RS256

I have fixed these in my last three commits, I also have included comments to evidence from where these changes came from, in term of normative references.

pyeudiw/tests/wallet_instance_attestations/test_attestation_request.py
manpace: Thanks for renaming the module to `wallet_attestations`. I had initially changed it from `wallet_instance_attestation `to `wallet_instance_attestations ` but the new naming is  definitely more appropriate.

